PR #1, opened 2016-02-04
Initial support for SaltStack.

geomacy:
See https://issues.apache.org/jira/browse/BROOKLYN-217.

This commit merges code from incubator-brooklyn, following the move to the new repos
after graduation. As such, it unfortunately loses the commit history, but the code
is based on the work done by Hadrian Zbarcea. See the Jira link above for history
and discussion.

NOTE this PR supersedes https://github.com/apache/brooklyn-server/pull/4, see the relevant [review comment](https://github.com/apache/brooklyn-server/pull/4#issuecomment-179360438).


geomacy:
Going to close/open to retry Jenkins.

geomacy:
Reopen to retry Jenkins build.

software/cm/salt/pom.xml
```diff
@@ -0,0 +1,89 @@
...
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.brooklyn</groupId>
+      <artifactId>brooklyn-software-base</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.brooklyn</groupId>
+      <artifactId>brooklyn-software-database</artifactId>
```
ahgittin: why does it depend on `database` ?

ahgittin:
Looks great, have merged:

* library: https://github.com/apache/brooklyn-library/pull/1
* docs: https://github.com/apache/brooklyn-docs/pull/9
* dist: https://github.com/apache/brooklyn-dist/pull/4

Added some minor cleanups at:
* library: https://github.com/apache/brooklyn-library/commit/d78795eeb1fc8f4d9da62c58e4723f4e899796fc
* docs: https://github.com/apache/brooklyn-docs/commit/a399f21de1d79ddf82610eddc6373653979a7bcb

In particular I removed `database` as a dependency, and removed causes of warnings.

Really nice work!  If Ansible is similar I'll be very happy /cc @ygy !

Next step I think should be to supply configuration.  Maybe the most natural way would be to define pillar *contents* of files inline, if pillars are the idiomatic way to give salt configuration.  For instance:

    pillarData:
      mysql/init.sls:
        # Manage databases
        database:
        - orders
        schema:
          orders:
            load: True
            source: http://myhost:8080/mysql/files/orders.schema

That would do enough to allow dependent configuration.  We'd probably afterwards want:
* a way to refer to "local" files (note I changed `source` to use an `http` url rather than a `salt:` url above) -- NB that's wanted across Brooklyn in yaml blueprints!
* better sensors (but not sure how)

Just my 2p on where this could go.  Inline config is the only thing which feels like it's missing, to me.

geomacy:
Thanks Alex, I'll add the above into my list of things to do.  I like that use of "mysql/init.sls" as a key, i.e. interpret the key as a piece of data (a file path) rather than just have it as a keyword.

PR #5, opened 2016-02-11
BROOKLYN-222 -  Ansible integration

ygy:
- Ansible install for Ubuntu and CentOS
- Playbooks can be specified via URL or directly as a YAML
- Support for custom start/stop commands
- Support for command to support setting a useful value for the
service.isUp sensor
- Generic check for a running service based on the provided serviceName
- Add effector for ad-hoc ansible commands.
- Addition of Ansible extra vars from Brooklyn config


geomacy:
For the sake of consistency I will suggest having a separate jar for the Ansible integration, just as there is for Salt, so have a new directory brooklyn-library/software/cm/ansible, with its own pom file, and move the code into it.  (This also means we would need to add an entry in brooklyn-dist/all/pom.xml for the new jar, alongside that for Salt).  I think I would rather have separate jars for each of the Ansible, Salt, Chef integrations than have them all in the same jar.

software/cm/src/main/java/org/apache/brooklyn/entity/cm/ansible/AnsibleEntityImpl.java
```diff
@@ -0,0 +1,61 @@
...
+        if (!Strings.isBlank(playbookName)) setDefaultDisplayName(playbookName + " (ansible)");
+
+        super.init();
+
+        lifecycleTasks = new AnsibleLifecycleEffectorTasks();
+
+        lifecycleTasks.attachLifecycleEffectors(this);
+    }
+
+    @Override
```
geomacy: Remove '@Override' from implementation of interface method. Also in line 49.

software/cm/src/main/java/org/apache/brooklyn/entity/cm/ansible/AnsibleLifecycleEffectorTasks.java
```diff
@@ -0,0 +1,203 @@
...
+        if (!result) {
+            LOG.warn("No way to check whether "+entity()+" is running; assuming yes");
+        }
+        entity().sensors().set(SoftwareProcess.SERVICE_UP, true);
+        
+        Maybe<SshMachineLocation> machine = Locations.findUniqueSshMachineLocation(entity().getLocations());
+
+        if (machine.isPresent()) {
+            
+            String serviceName = String.format("[%s]%s", entity().config().get(SERVICE_NAME).substring(0, 1),
```
geomacy: might be worth a one liner comment in here to explain why this format is used.

software/cm/src/main/java/org/apache/brooklyn/entity/cm/ansible/AnsibleLifecycleEffectorTasks.java
```diff
@@ -0,0 +1,203 @@
...
+            LOG.warn("Location(s) {} not an ssh-machine location, so not polling for status; setting serviceUp immediately", entity().getLocations());
+        }
+    }
+
+    protected boolean tryCheckStartService() {
+        if (getServiceName()==null) return false;
+
+        // if it's still up after 5s assume we are good (default behaviour)
+        Time.sleep(Duration.FIVE_SECONDS);
+        if (!((Integer)0).equals(DynamicTasks.queue(SshEffectorTasks.ssh(String.format(entity().config().get(AnsibleConfig.ANSIBLE_SERVICE_START), getServiceName()))).get())) {
```
geomacy: For legibility (not least on github) can we wrap these lines at about 120 characters?

software/cm/src/main/java/org/apache/brooklyn/entity/cm/ansible/AnsibleLifecycleEffectorTasks.java
```diff
@@ -0,0 +1,203 @@
...
+        if (machine.isPresent()) {
+            
+            String serviceName = String.format("[%s]%s", entity().config().get(SERVICE_NAME).substring(0, 1),
+                    entity().config().get(SERVICE_NAME).substring(1));
+            String checkCmd = String.format("ps -ef | grep %s", serviceName);
+
+            Integer serviceCheckPort = entity().config().get(ANSIBLE_SERVICE_CHECK_PORT);
+
+            if (serviceCheckPort != null) {
+                checkCmd = String.format("sudo ansible localhost -c local -m wait_for -a \"host=0.0.0.0 port=%d\"", serviceCheckPort);
```
geomacy: Would it be worth having a config key for the interface, with a default of "0.0.0.0"?  I guess it's unlikely to be needed much but maybe sometimes?

software/cm/src/main/java/org/apache/brooklyn/entity/cm/ansible/AnsiblePlaybookTasks.java
```diff
@@ -0,0 +1,109 @@
...
+        Entity entity = EffectorTasks.findEntity();
+        String yaml = entity.config().get(AnsibleConfig.ANSIBLE_PLAYBOOK_YAML);
+
+        return Tasks.sequential("build ansible playbook file for "+ playbook,
+            SshEffectorTasks.put(Urls.mergePaths(ansibleDirectory) + "/" + playbook + ".yaml")
+                .contents(yaml).createDirectory());
+    }
+
+    public static TaskFactory<?> runAnsible(final String dir, Object extraVars, String playbookName) {
+        String cmd = String.format("sudo ansible-playbook -i \"localhost,\" -c local "
```
geomacy: Maybe better to wrap this in BashCommands.sudo("ansible-playbook ...") rather than have the sudo in the command.

software/cm/src/main/java/org/apache/brooklyn/entity/cm/ansible/AnsibleLifecycleEffectorTasks.java
```diff
@@ -0,0 +1,203 @@
...
+        return entity().config().get(ANSIBLE_PLAYBOOK);
+    }
+
+    protected void startWithAnsibleAsync() {
+        String installDir = Urls.mergePaths(getBaseDir(), "installs/ansible");
+
+        String playbookUrl = entity().config().get(ANSIBLE_PLAYBOOK_URL);
+        String playbookYaml = entity().config().get(ANSIBLE_PLAYBOOK_YAML);
+
+        if (Strings.isNonBlank(playbookUrl) && Strings.isNonBlank(playbookYaml)) {
```
geomacy: Worth also adding a check for the case where both of them are blank, as otherwise the statements just below will run ansible without having established a playbook by either means.

software/cm/src/main/java/org/apache/brooklyn/entity/cm/ansible/AnsibleLifecycleEffectorTasks.java
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    @Override
+    protected String startProcessesAtMachine(Supplier<MachineLocation> machineS) {
+        startWithAnsibleAsync();
+
+        return "ansible start tasks submitted";
+    }
+
+    protected String getPlaybookName() {
```
geomacy: Do we really need playbookName?  In the case of a download we could infer it from the URL, and in the case of inline YAML we could make it up.  Actually we could make it up even in the case of download from URL.  Removing it would reduce the number of mandatory things the blueprint author has to think about from three to two (serviceName, plus playbook content).

geomacy:
Some few comments above, all minor in the sense that none actually _needs_ to be done.  Otherwise looks great to me.

ygy:
@geomacy I've addressed your comments

geomacy:
That all looks great @ygy, thanks.  As far as I am concerned this should be merged.

geomacy:
Looks good to me.

PR #6, opened 2016-02-15
add ability to pass DB creation script as a url to a freemarker template

johnmccabe:
Added to `DatastoreMixins` as part of the `CanGiveCreationScript` interface, should be available to all DBs, have tested with PostgreSQL and MySQL.

Example:
```
name: DB Test
location: byon1
services:
  id: db
  name: MySQL DB
  brooklyn.config:
    daytrader.dbname: mydbname
    daytrader.username: mydbuser
    daytrader.password: mysecurepassword
    datastore.creation.script.template.url: "https://myhost.domain.com/templates/mydbname.template.ddl"
```

With the template DDL using standard Freemarker markup.

### DIFF
--- a/software/database/src/main/java/org/apache/brooklyn/entity/database/DatastoreMixins.java
+++ b/software/database/src/main/java/org/apache/brooklyn/entity/database/DatastoreMixins.java
@@ -18,19 +18,25 @@
  */
 package org.apache.brooklyn.entity.database;
 
+import java.io.ByteArrayInputStream;
 import java.io.InputStream;
+import java.io.UnsupportedEncodingException;
 
 import javax.annotation.Nullable;
 
+import com.google.common.collect.ImmutableMap;
 import org.apache.brooklyn.api.effector.Effector;
 import org.apache.brooklyn.api.entity.Entity;
 import org.apache.brooklyn.api.sensor.AttributeSensor;
 import org.apache.brooklyn.config.ConfigKey;
 import org.apache.brooklyn.core.config.ConfigKeys;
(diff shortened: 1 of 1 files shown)


software/database/src/main/java/org/apache/brooklyn/entity/database/mysql/MySqlSshDriver.java
```diff
@@ -220,6 +220,10 @@ protected void copyDatabaseConfigScript() {
 
     protected boolean copyDatabaseCreationScript() {
         String creationScriptContents = DatastoreMixins.getDatabaseCreationScriptAsString(entity);
+
+        if (creationScriptContents==null)
+            creationScriptContents = processTemplate(entity.getConfig(MySqlNode.TEMPLATE_CREATION_SCRIPT));
```
neykov: Why not move it a level up in `DatastoreMixins.getDatabaseCreationScriptAsString` so all DBs can take advantage of the change. Otherwise looks good.

johnmccabe: Yup @ahgittin suggested the same, just testing the fix at the mo. Thanks @neykov

johnmccabe:
@neykov I've added it to the `DatastoreMixins` (updated the title/description for the PR)

neykov:
Thanks @johnmccabe looks great.
Can you squash, remove MySql* files changes, then good to merge.

johnmccabe:
updated, thanks @neykov

PR #7, opened 2016-02-17
MS SQL installation - remove passing Administrator credentials

bostko:
Without this PR it installation on Amazon tells something like:
"Not allowed to authenticate as Administrator" for the last command

However now installation ends with "Exit Code -2068774911".
If I launch it again for the second time with plain winrm4j tool it works.
I will try with a BYON location and see how it installs from there.


ahgittin:
@bostko should this be merged or is it still WIP?  the description makes it sounds like one error is replaced with another error.

bostko:
@ahgittin One error is fixed,
But then continuing the installation it fails on the end.

bostko:
@ahgittin now it works.

PR #8, opened 2016-02-18
Feature/paas cf support

kiuby88:
This commit depends on [brooklyn-server#30](https://github.com/apache/brooklyn-server/pull/30)
Adding PaaS CloudFoundry support to `TomcatServer` entity.

The drivers for CloudFoundry are added to `SoftwareProcess` entity and `JavaWebAppSoftwareProcess` implementation entity. 
The final PaaS support is added to `TomcatServer` entity using a new driver which is focused on `CloudFoundryPaaSLocation`.

Then, `TomcatServer` can be started in a Ssh location or in a CloudFoundryLocation.

For example, the following `hello-world-sql` application has a TomcatServer.

```
name: appserver-web-aws
services:
- type: org.apache.brooklyn.entity.webapp.tomcat.TomcatServer
  name: AppServer HelloWorld
  location: <LOCATION>
  brooklyn.config:
    wars.root: https://github.com/kiuby88/brooklyn-library/releases/download/hello-world-sql-paas/brooklyn-example-hello-world-sql-webapp-in-paas.war
    http.port: 8080+
    java.sysprops: 
      brooklyn.example.db.url: $brooklyn:formatString("jdbc:%s%s?user=%s&password=%s",
         component("db").attributeWhenReady("datastore.url"), "visitors", "brooklyn", "br00k11n")

  id: db
  location: aws-ec2:us-west-2
  name: DB HelloWorld Visitors
  brooklyn.config:
    datastore.creation.script.url: https://raw.githubusercontent.com/apache/incubator-brooklyn/master/brooklyn-server/launcher/src/test/resources/visitors-creation-script.sql
```

`TomcatServer`'s location can contains a Ssh- or CloudFoundry-based location. If `<LOCATION>` is equals to e.g., `aws-ec2:us-west-2, TomcatServer will be deployed on AWS Oregon.
But, it <LOCATION> is a CloudFoundry-based location, for example [Pivotal](https://run.pivotal.io/):

```
  location:
    cloudfoundry:
      user: <user>
      password: <pass>
      org: <org>
      endpoint: https://api.run.pivotal.io
      space: <space>
      address: run.pivotal.io
```

`TomcatServer` entity will be deployed using Pivotal Web Services.

Please, note that the yaml is the same for deploying the application on Ssh or CloudFoundry. 
#### Tested Providers

This approach has been tested using:
- [Pivotal](https://run.pivotal.io/)
- [Bluemix](https://console.ng.bluemix.net/)
#### RoadMap

This is a first approach. If it is accepted a lot of entities could be enabled to operate on CloudFoundry.

### DIFF
--- a/examples/webapps/hello-world-sql/src/main/webapp/db.jsp
+++ b/examples/webapps/hello-world-sql/src/main/webapp/db.jsp
@@ -39,6 +39,11 @@ deployed by brooklyn, to show <b>SQL database interactivity</b>.
 
 <%
 String url=System.getProperty("brooklyn.example.db.url");
+
+if((url == null) || (url.isEmpty())) {
+    url = System.getenv("brooklyn.example.db.url");
+}
+
 //URL should be supplied e.g. ""-Dbrooklyn.example.db.url=jdbc:mysql://localhost/visitors?user=brooklyn&password=br00k11n"
 //(note quoting needed due to ampersand)
 if (url==null) {
--- a/software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/couchbase/CouchbaseNodeImpl.java
+++ b/software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/couchbase/CouchbaseNodeImpl.java
@@ -26,6 +26,7 @@ import java.util.Map;
 import java.util.Set;
 
 import org.apache.brooklyn.api.entity.Entity;
+import org.apache.brooklyn.api.location.Location;
 import org.apache.brooklyn.api.location.MachineProvisioningLocation;
 import org.apache.brooklyn.api.sensor.AttributeSensor;
 import org.apache.brooklyn.api.sensor.SensorEvent;
@@ -107,8 +108,8 @@ public class CouchbaseNodeImpl extends SoftwareProcessImpl implements CouchbaseN
         });
     }
 
-    protected Map<String, Object> obtainProvisioningFlags(@SuppressWarnings("rawtypes") MachineProvisioningLocation location) {
-        ConfigBag result = ConfigBag.newInstance(super.obtainProvisioningFlags(location));
+    protected Map<String, Object> obtainFlagsForLocation(@SuppressWarnings("rawtypes") Location location) {
+        ConfigBag result = ConfigBag.newInstance(super.obtainFlagsForLocation(location));
         result.configure(CloudLocationConfig.OS_64_BIT, true);
         return result.getAllConfig();
     }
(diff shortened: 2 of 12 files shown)


robertgmoss:
@kiuby88 I'm not convinced it is `TomcatServer` that needs to be deployed to CF.  I think it makes more sense to have a 'CloudFoundryWebApp' which can have a manifest as config.  WDYT?

kiuby88:
@robertgmoss thanks a lot for your comment.
 It makes sense and it should be very easy to implement. But manifest, it may be very closed to the CloudFoundry web console. I do not know.

But, I wanted to try to start an entity in SSH location and PaaS location depending on the location. So, you have an agnostic yaml plan (blueprint) that you can started on SSH or PaaS easly, it was my idea.
In any case, [TomcatCloudFoundryDriver](https://github.com/kiuby88/brooklyn-library/blob/feature/paas-cf-support/software/webapp/src/main/java/org/apache/brooklyn/entity/webapp/tomcat/TomcatCloudFoundryDriver.java) is very simple because all work is carried out by  `JavaWebAppCloudFoundryDriver`. 
So, specific driver, as `TomcatCloudFoundryDriver` or an future `JBossCloudFoundryDriver` will focus on provide the required buildpack (e.g., [JBoss Buildpack](https://github.com/cloudfoundry-community/jboss-buildpack)) to `JavaWebAppCloudFoundryDriver`. 

In any case, it was just only my opinion and I can try your proposal.

ahgittin:
Awesome @kiuby88 !

However I tend to agree with @robertgmoss this would be better implemented as an entity than a driver for `TomcatServer`.  The mismatches I'm thinking of are:
* with CF I can create several instances, and I'm given a route which goes to any of them -- so it's functionally more similar to a `ControlledDynamicWebAppCluster`
* `TomcatServer` expects to be able to configure Tomcat and deploy multiple WARs
* the driver normally kicks in once a machine is provisioned, which isn't the case here

My inclination would be an `abstract CloudFoundryAppForBuildPack` entity with a `CloudFoundryTomcatApp` child.  There is probably no need for a driver so implementation of that should be fairly easy as you note given what you've done, mainly copying some of what `TomcatServer` does moving existing code around.  Nice if we can loop in some of the existing sensor names from `WebAppCluster`.

On `CloudFoundryAppForBuildPack` we can also expose things like the number of instances and the route and download of items -- which would make other items like `CloudFoundryRubyApp` and `CloudFoundryJBossApp` also straightforward.

BTW it's very cool what you've done being able to take a DB and configure that against CF directly!

andreaturli:
@kiuby88 thanks again for your effort. Same comment made at https://github.com/apache/brooklyn-server/pull/30#issuecomment-289792727 for this PR as well. Would you close this and possibly re-open a new PR, even to a different repo, in case needed?

kiuby88:
Sure @andreaturli.
Let me close the PRs. I will always be available to update and re-open these PRs if it is necessary

PR #9, opened 2016-02-18
Validation of blueprint projects

iyovcheva:
Every time `CatalogValidationTest ` is being run it will validate the manually set `REPO_URL` which is now `https://github.com/brooklyncentral/brooklyn-community-catalog`. If there is a new catalog added to `brooklyn-community-catalog`, which cannot be successfully validated, the test will fail with message: 
*"\<catalog-name/ur\> is not valid catalog"*. 

To validate a new catalog before adding it to `brooklyn-community-catalog` one can run only `CatalogValidationTest.testSingleCatalogValidation()` and change `VALID_CTALOG_URL` with its url.

### DIFF
--- a/brooklyn-validations/pom.xml
+++ b/brooklyn-validations/pom.xml
@@ -1,0 +1,103 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
+    distributed with this work for additional information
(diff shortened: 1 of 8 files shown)


iyovcheva:
test that please

iyovcheva:
retest this please

ahgittin:
Does this belong here or as part of the catalog project? /cc @tbouron

tbouron:
Hum, I don't think so. It belongs to the community catalog, especially if we are going for the Github integration for devs

PR #10, opened 2016-02-19
Update abstractNonProvisionedController (for aws elb)

aledsage:
See https://github.com/cloudsoft/brooklyn-aws-elb/ for an example usage of this. That is not in Brooklyn core because it depends on the aws-sdk, which would pull in > 10MB of additional dependencies.

### DIFF
--- a/software/webapp/src/main/java/org/apache/brooklyn/entity/proxy/AbstractNonProvisionedControllerImpl.java
+++ b/software/webapp/src/main/java/org/apache/brooklyn/entity/proxy/AbstractNonProvisionedControllerImpl.java
@@ -18,9 +18,9 @@
  */
 package org.apache.brooklyn.entity.proxy;
 
-import static com.google.common.base.Preconditions.checkState;
 import static org.apache.brooklyn.util.JavaGroovyEquivalents.groovyTruth;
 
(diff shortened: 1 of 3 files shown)


software/webapp/src/main/java/org/apache/brooklyn/entity/proxy/AbstractNonProvisionedControllerImpl.java
```diff
@@ -96,96 +109,180 @@ public boolean isActive() {
...
+        removeServerPoolMemberTrackingPolicy();
     }
-        
-    protected void addServerPoolMemberTrackerPolicy() {
+
+    protected void doStop() {
+    }
+    
+    protected void postStop() {
+        removeServerPoolMemberTrackingPolicy();
```
neykov: Are you calling it in on purpose in both `preStop` & `postStop`?

ahgittin:
LGTM.  agree @neykov's comment needs addressing, also the constructors change will have tiny conflict w #11 as you noted, but otherwise good to merge.

aledsage:
Thanks all - changes incorporated, jenkins build confirmed; merging now.

PR #11, opened 2016-02-25
fix failing test by not destroying app on stop

ahgittin:
and minor other tidies

### DIFF
--- a/software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
+++ b/software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
@@ -22,13 +22,13 @@ import static java.lang.String.format;
 import static org.apache.brooklyn.util.ssh.BashCommands.INSTALL_CURL;
 import static org.apache.brooklyn.util.ssh.BashCommands.INSTALL_TAR;
 import static org.apache.brooklyn.util.ssh.BashCommands.addSbinPathCommand;
-import static org.apache.brooklyn.util.ssh.BashCommands.sbinPath;
 import static org.apache.brooklyn.util.ssh.BashCommands.alternatives;
 import static org.apache.brooklyn.util.ssh.BashCommands.chainGroup;
 import static org.apache.brooklyn.util.ssh.BashCommands.commandToDownloadUrlAs;
 import static org.apache.brooklyn.util.ssh.BashCommands.ifExecutableElse;
 import static org.apache.brooklyn.util.ssh.BashCommands.ifNotExecutable;
 import static org.apache.brooklyn.util.ssh.BashCommands.ok;
+import static org.apache.brooklyn.util.ssh.BashCommands.sbinPath;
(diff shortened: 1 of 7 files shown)


software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
```diff
@@ -108,9 +108,10 @@ public void preInstall() {
 
     @Override
     public void install() {
-        if (entity.getConfig(Attributes.DOWNLOAD_URL) != null) {
-            LOG.warn("Ignoring download.url {}, use download.url.rhelcentos or download.url.mac", entity.getConfig(Attributes.DOWNLOAD_URL));
-        }
+        // TODO download URL can be inferred
+//        if (entity.getConfig(Attributes.DOWNLOAD_URL) != null) {
```
aledsage: Under what circumstances would we uncomment this code? Should we just delete it?
Is the TODO saying that we should infer it, or that the warning was pointless because we do infer it?

software/webapp/src/main/java/org/apache/brooklyn/entity/proxy/AbstractNonProvisionedControllerImpl.java
```diff
@@ -57,19 +57,19 @@
     public AbstractNonProvisionedControllerImpl() {
         this(MutableMap.of(), null, null);
     }
-    public AbstractNonProvisionedControllerImpl(Map properties) {
+    public AbstractNonProvisionedControllerImpl(Map<?,?> properties) {
```
aledsage: In https://github.com/apache/brooklyn-library/pull/10 (not merged or reviewed yet by anyone), I juste deleted these constructors.

software/webapp/src/test/java/org/apache/brooklyn/entity/proxy/TrackingAbstractControllerImpl.java
```diff
@@ -56,7 +55,7 @@ protected void reconfigureService() {
     }
 
     @Override
-    public Class getDriverInterface() {
+    public Class<?> getDriverInterface() {
```
aledsage: For some reason I thought this didn't work (i.e. we couldn't use `Class<?>`). Maybe because the supertype `DriverDependentEntity<D extends EntityDriver>` declares it as `Class<D> getDriverInterface();` and `Class<?>` is not a subtype of `Class<D>`?

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
ahgittin: good spot, should be removed

software/webapp/src/test/java/org/apache/brooklyn/entity/proxy/TrackingAbstractControllerImpl.java
ahgittin: not sure why but it seems to allow this here

PR #13, opened 2016-03-02
MS SQL improvements

bostko:
- catch possible failures on each step from the install

### DIFF
--- a/software/database/src/main/resources/org/apache/brooklyn/entity/database/mssql/installmssql.ps1
+++ b/software/database/src/main/resources/org/apache/brooklyn/entity/database/mssql/installmssql.ps1
@@ -24,7 +24,21 @@ $Path = "C:\sql2008.iso"
 $Username = "${config['mssql.download.user']}"
 $Password = '${config['mssql.download.password']}'
 
+& winrm set winrm/config/service/auth '@{CredSSP="true"}'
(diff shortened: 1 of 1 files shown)


software/database/src/main/resources/org/apache/brooklyn/entity/database/mssql/installmssql.ps1
```diff
@@ -49,5 +54,9 @@ $exitCode = Invoke-Command -ComputerName $env:COMPUTERNAME -Credential $mycreds
...
     $process.ExitCode
 } -Authentication CredSSP -ArgumentList $driveLetter
-
-exit $exitCode
+} Catch {
+ Write-Error $_.Exception
+ Write-Host 'Exception logged'
+ exit 1
+}
+exit $exitCode
```
neykov: is `$exitCode` available out of the try-catch scope?

bostko: @neykov good point.
It works.
```powershell
Write-Host hi
Try {
$b = 3434
} Catch {
 Write-Host 'test'
}
Write-Host $b # prints 3434
```
I think that at least from closure prospective PowerShell should be user friendly :)

PR #15, opened 2016-03-04
Define MAIN_URI inside SolrServer

bostko:
I made this change in order to be accessible when listing the static fields of SolrServer for other extensions, like transformers.
@aledsage Can you review


ahgittin:
LGTM

PR #19, opened 2016-03-11
Add creation of /etc/ansible hosts during install.

geomacy:
This allows replacement of the command form

   ansible-playbook -i \"localhost,\" -c local  whatever

 with

   ansible-playbook whatever

 and also supports the possibility that playbooks can modify the hosts file e.g. by adding additional groups that included roles require.

 For example the following playbook modifies the hosts to add a tomcat-server group, in order to install an example ansible from the github examples repo.
```yaml
 name: multi
 location:
   red1
 services:
 - serviceType: brooklyn.entity.basic.SameServerEntity
   name: Entities
   brooklyn.children:

   - serviceType: org.apache.brooklyn.entity.cm.ansible.AnsibleEntity
     id: bootstrap
     service.name: crond
     playbook: bootstrap
     playbook.yaml: |
         ---
         - hosts: localhost
           tasks:
           - shell: printf "[tomcat-servers]\nlocalhost ansible_connection=local\n" >> /etc/ansible/hosts
           - file: path=/etc/ansible/playbooks state=directory mode=0755
           - get_url: url=https://github.com/ansible/ansible-examples/archive/master.zip dest=/tmp/master.zip mode=0440
           - command: unzip -o -d /etc/ansible/playbooks /tmp/master.zip

   - serviceType: org.apache.brooklyn.entity.stock.BasicApplication
     start.latch: $brooklyn:component("bootstrap").attributeWhenReady("service.isUp")
     brooklyn.children:
     - type: org.apache.brooklyn.entity.cm.ansible.AnsibleEntity
       name: test
       service.name: tomcat
       playbook: tomcat
       playbook.yaml: |
           ---
           - hosts: localhost
           - include: /etc/ansible/playbooks/ansible-examples-master/tomcat-standalone/site.yml
             vars:
                 http_port: 8080
                 https_port: 8443
                 admin_username: admin
                 admin_password: secret
```

### DIFF
--- a/software/cm/ansible/src/main/java/org/apache/brooklyn/entity/cm/ansible/AnsiblePlaybookTasks.java
+++ b/software/cm/ansible/src/main/java/org/apache/brooklyn/entity/cm/ansible/AnsiblePlaybookTasks.java
@@ -19,10 +19,10 @@
 package org.apache.brooklyn.entity.cm.ansible;
 
 import org.apache.brooklyn.api.entity.Entity;
-import org.apache.brooklyn.api.mgmt.TaskAdaptable;
 import org.apache.brooklyn.api.mgmt.TaskFactory;
 import org.apache.brooklyn.core.effector.EffectorTasks;
 import org.apache.brooklyn.core.effector.ssh.SshEffectorTasks;
+import org.apache.brooklyn.util.collections.MutableList;
 import org.apache.brooklyn.util.core.ResourceUtils;
 import org.apache.brooklyn.util.core.task.Tasks;
 import org.apache.brooklyn.util.core.task.system.ProcessTaskFactory;
@@ -35,7 +35,6 @@ import org.yaml.snakeyaml.DumperOptions;
 import org.yaml.snakeyaml.Yaml;
 
 import static org.apache.brooklyn.core.effector.ssh.SshEffectorTasks.ssh;
-import static org.apache.brooklyn.util.ssh.BashCommands.pipeTextToFile;
 import static org.apache.brooklyn.util.ssh.BashCommands.sudo;
 
 public class AnsiblePlaybookTasks {
@@ -110,7 +109,25 @@ public class AnsiblePlaybookTasks {
     }
 
     public static TaskFactory<?> setUpHostsFile(String installDir, boolean force) {
-        return ssh(pipeTextToFile("localhost ansible_connection=local", "/etc/ansible/hosts"))
-            .summary("write hosts file");    }
+        final String hostsFile = "/etc/ansible/hosts";
+        final String tempFile = "/tmp/hosts";
+        String checkInstalled = !force ? "which ansible || " : "";
+        return sshCommands(
+                checkInstalled + " {",
+                sudo("rm -f " + hostsFile),
+                "echo 'localhost ansible_connection=local' > " + tempFile,
+                sudo("mv " + tempFile + " " + hostsFile),
+                "}"
+            )
+            .requiringExitCodeZero()
+            .summary("write hosts file");
+    }
+
+
+    public static SshEffectorTasks.SshEffectorTaskFactory<Integer> sshCommands(String line, String... lines) {
+        final MutableList.Builder<String> builder = MutableList.<String>builder().add(line);
+        builder.addAll(lines);
+        return ssh(Strings.join(builder.build(), "\n"));
+    }
 }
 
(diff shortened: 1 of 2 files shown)


software/cm/ansible/src/main/java/org/apache/brooklyn/entity/cm/ansible/AnsiblePlaybookTasks.java
```diff
@@ -105,5 +107,27 @@ private static String optionalExtraVarsParameter(Object extraVars) {
         }
         return " --extra-vars \"@" + EXTRA_VARS_FILENAME + "\" ";
     }
+
+    public static TaskFactory<?> setUpHostsFile(String installDir, boolean force) {
+        final String hostsFile = "/etc/ansible/hosts";
+        final String tempFile = "/tmp/hosts";
```
neykov: Could add a random suffix from `Identifiers.makeRandomId` just in case.

software/cm/ansible/src/main/java/org/apache/brooklyn/entity/cm/ansible/AnsiblePlaybookTasks.java
```diff
@@ -105,5 +107,27 @@ private static String optionalExtraVarsParameter(Object extraVars) {
...
+        String checkInstalled = !force ? "which ansible || " : "";
+        return sshCommands(
```
neykov: Should this be `sudo`'ed?

software/cm/ansible/src/main/java/org/apache/brooklyn/entity/cm/ansible/AnsiblePlaybookTasks.java
```diff
@@ -105,5 +107,27 @@ private static String optionalExtraVarsParameter(Object extraVars) {
...
+                checkInstalled + " {",
+                sudo("rm -f " + hostsFile),
+                "echo 'localhost ansible_connection=local' > " + tempFile,
```
neykov: Could send this directly to `/etc/ansible/hosts`. Should be atomic? Then the wrapper block won't be needed.

software/cm/ansible/src/main/java/org/apache/brooklyn/entity/cm/ansible/AnsiblePlaybookTasks.java
```diff
@@ -105,5 +107,27 @@ private static String optionalExtraVarsParameter(Object extraVars) {
         }
         return " --extra-vars \"@" + EXTRA_VARS_FILENAME + "\" ";
     }
+
+    public static TaskFactory<?> setUpHostsFile(String installDir, boolean force) {
+        final String hostsFile = "/etc/ansible/hosts";
+        final String tempFile = "/tmp/hosts";
+        String checkInstalled = !force ? "which ansible || " : "";
```
neykov: Should be `&&` so the wrapper block is executed only if ansible is found?

neykov:
Mostly minor comments. Can be merged once `&&` vs `||` usage is confirmed.

software/cm/ansible/src/main/java/org/apache/brooklyn/entity/cm/ansible/AnsiblePlaybookTasks.java
geomacy: I needed to write it as sudo, but when you try to do the create directly under sudo with a redirect, it breaks (redirects don't work under sudo).  The operation is still atomic really, it is the "mv"  - the tempFile is just a convenience.

neykov: For redirects under sudo you can use `sh -c 'echo XXX > /tmp/out'`

ahgittin: my favourite is

    echo XXX | sudo tee /etc/privileged_file

or in java

    "echo XXX | "+sudo("tee /etc/privileged_file")

with `tee -a` to append

geomacy:
have added an update in response to the comments.  Logic of setUpHostsFile at line 112 is "if you don't want to force the write, then check if the required value is there first, else write it regardless".

neykov:
LGTM

aledsage:
LGTM; merging.

PR #23, opened 2016-03-23
Fix: ElasticSearch advertise datastore url

aledsage:
The first commit is the important one; the second is cosmetic, to get rid of deprecation warnings in the tests.


bostko:
Looks good!

aledsage:
Jenkins build failure looks like yet another unrelated strange environment thing?

```
[FATAL] Non-resolvable parent POM: Could not find artifact org.apache.brooklyn:brooklyn-parent:pom:0.9.0-SNAPSHOT and 'parent.relativePath' points at wrong local POM @ line 24, column 13
```

PR #25, opened 2016-03-24
Populate catalog from bundle bom

geomacy:
The PR adds updates to the existing jars to add a catalog.bom file that includes the details of the various entities in each.  

See https://github.com/apache/brooklyn-server/pull/80 for details; this PR depends on that one and should be merged after it.

### DIFF
--- a/software/webapp/src/main/resources/catalog.bom
+++ b/software/webapp/src/main/resources/catalog.bom
@@ -1,0 +1,234 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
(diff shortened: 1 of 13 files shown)


neykov:
@geomacy Do you think we should keep `scanJavaAnnotations` even if not supported? It's spamming the logs for the existing .boms that are already merged.

geomacy:
That's a good thought, will do that update.

software/webapp/src/main/resources/catalog.bom
```diff
@@ -0,0 +1,234 @@
...
+      description: |
+        Create a load balancer which will point at members in the group entity
+        referred to by the config key "serverPool".
+        The sensor advertising the port can be configured with the "member.sensor.portNumber" config key,
+        defaulting to `http.port`; all member entities which have published "service.up" will then be picked up.
+      item:
+        type: org.apache.brooklyn.entity.proxy.nginx.NginxController
+        name: Load Balancer (nginx)
+
+    - id: resilient-bash-web-cluster-template
```
neykov: Is this duplicating `default.catalog.bom` from the dist on purpose? 
Would have been nice if we supported catalog inclusions.

neykov:
Can you squash.

software/webapp/src/main/resources/catalog.bom
geomacy: It is done on purpose, for the moment - it would be nice to have catalog inclusions, but as far as I'm aware there's nothing available at the moment.  If and when we move to only running on Karaf then the default.catalog.bom can go away, and each bundle can have its own example applications like this.

geomacy:
Changes squashed.

PR #26, opened 2016-03-25
Fix: ElasticSearch used Java version

ygy:
Making ElasticSearch to use JavaSoftwareProcessDriver so the correct
java version is installed. ElasticSearch used to install java via
BashCommands#installJavaLatestOrWarn(). It prevents setting the
installed java version as default when:
- the VM image comes with preinstalled jdk 6 or jdk7
- CentOS version of JDK8 doesn't update the default automatically


aledsage:
@ygy can you explain how this fixes it? Under the covers, it looks like it will still just call `BashCommands.installJava(...)` which will do the `yum install` but will not update the CentOS `alternatives` to point at the new Java.

Or does `JavaSoftwareProcessSshDriver` do something else to get the right (newer) java on the path?

aledsage:
Have chatted with @ygy - the changes mean we install Java 7 rather than Java 8. It seems that `yum install` for Java 7 will ensure it is at the head of the path, but doing the same for Java 8 will not.

Confirmed it worked with the `us-west-2/ami-56e48766` (which it previously failed with).

Merging.

PR #32, opened 2016-04-19
CouchDB installation fixes

ygy:
- CouchDB is installed from source now
- CouchDB dependencies are installed by the official repos
- Fixed SuSE Linux support

### DIFF
--- a/software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/couchdb/CouchDBNode.java
+++ b/software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/couchdb/CouchDBNode.java
@@ -40,7 +40,7 @@ public interface CouchDBNode extends SoftwareProcess, WebAppService {
 
     @SetFromFlag("downloadUrl")
     AttributeSensorAndConfigKey<String, String> DOWNLOAD_URL = ConfigKeys.newSensorAndConfigKeyWithDefault(SoftwareProcess.DOWNLOAD_URL,
-            "http://www-eu.apache.org/dist/couchdb/source/${version}/apache-couchdb-${version}.tar.gz");
+            "http://www.mirrorservice.org/sites/ftp.apache.org/couchdb/source/${version}/apache-couchdb-${version}.tar.gz");
 
     @SetFromFlag("archiveNameFormat")
     ConfigKey<String> ARCHIVE_DIRECTORY_NAME_FORMAT = ConfigKeys.newConfigKeyWithDefault(SoftwareProcess.ARCHIVE_DIRECTORY_NAME_FORMAT, "apache-couchdb-%s");
--- a/software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/couchdb/CouchDBNode.java
+++ b/software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/couchdb/CouchDBNode.java
@@ -22,6 +22,7 @@ import org.apache.brooklyn.api.catalog.Catalog;
 import org.apache.brooklyn.api.entity.ImplementedBy;
 import org.apache.brooklyn.config.ConfigKey;
 import org.apache.brooklyn.core.config.ConfigKeys;
+import org.apache.brooklyn.core.sensor.AttributeSensorAndConfigKey;
 import org.apache.brooklyn.core.sensor.BasicAttributeSensorAndConfigKey;
 import org.apache.brooklyn.entity.software.base.SoftwareProcess;
 import org.apache.brooklyn.entity.webapp.WebAppService;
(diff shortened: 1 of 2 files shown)


software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/couchdb/CouchDBNode.java
```diff
@@ -37,8 +38,15 @@
 @ImplementedBy(CouchDBNodeImpl.class)
 public interface CouchDBNode extends SoftwareProcess, WebAppService {
 
+    @SetFromFlag("downloadUrl")
+    AttributeSensorAndConfigKey<String, String> DOWNLOAD_URL = ConfigKeys.newSensorAndConfigKeyWithDefault(SoftwareProcess.DOWNLOAD_URL,
+            "http://www-eu.apache.org/dist/couchdb/source/${version}/apache-couchdb-${version}.tar.gz");
```
neykov: From Apache site: `Please use the backup mirrors only to download PGP and MD5 signatures to verify your downloads or if no other mirrors are working.`

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/couchdb/CouchDBNodeSshDriver.java
```diff
@@ -76,26 +82,85 @@ public CouchDBNodeSshDriver(CouchDBNodeImpl entity, SshMachineLocation machine)
...
+       cmds.add(BashCommands.installPackage(installGccPackageFlags, "couchdb-prerequisites-gcc"));
+       cmds.add(BashCommands.installPackage(installMakePackageFlags, "couchdb-prerequisites-make"));
+       cmds.add(BashCommands.installPackage(installPackageFlags, "couchdb-prerequisites"));
+       cmds.addAll(BashCommands.commandsToDownloadUrlsAs(couchdbUrls, coudhdbSaveAs));
+
+       cmds.add(format("tar xvzf %s", coudhdbSaveAs));
+       cmds.add(format("cd %s", getExpandedInstallDir()));
+
+       StringBuilder configureCommand = new StringBuilder("./configure")
+               .append(format(" --prefix=%s", getRunDir()))
```
neykov: Would it make more sense to install it to `getInstallDir()`?

neykov:
Looks good apart for a couple of minor comments.

ygy:
Thanks @neykov 
I've addressed the first comment and changed the download link.

I'm using the `runDir()` instead `getInstallDir()` to achieve the same result as in some other entities which use source base install, e.g. nginx

neykov:
Could using `runDir()` lead to unexpected results when deploying to non-pristine locations like `byon`? It will see that the install directory already exists, but create a clean run directory which doesn't contain the binaries.

ygy:
@neykov you are correct
Addressed your comments

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/couchdb/CouchDBNodeSshDriver.java
```diff
@@ -129,7 +199,7 @@ public void customize() {
     public void launch() {
         log.info("Launching  {}", entity);
         newScript(MutableMap.of(USE_PID_FILE, false), LAUNCHING)
-                .body.append(sudo(String.format("nohup couchdb -p %s -a %s -o couchdb-console.log -e couchdb-error.log -b &", getPidFile(), Os.mergePathsUnix(getRunDir(), getCouchDBConfigFileName()))))
+                .body.append(String.format("nohup ./bin/couchdb -p %s -a %s -o couchdb-console.log -e couchdb-error.log -b &", getPidFile(), Os.mergePathsUnix(getRunDir(), getCouchDBConfigFileName())))
```
kiuby88: The bin folder string is used in some different parts. Would it make sense to wrap `/bin/couchdb` folder with a method? 
Moreover, I looks that couchdb files are installed over different folders:
`/usr/bin/couchdb` contains the launch ssh file, and `/etc/couchdb` contains some configuration files  and `/usr/share/couchdb` contains the source to execute the server. Then, would it make sense try to store these paths in any variable or sensor? It looks that these values are concrete values of this driver, but  may it be useful for future improvement?

kiuby88:
Thanks a lot @ygy.
I am trying to understand better the drivers and I just wrote a pair of questions.

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/couchdb/CouchDBNodeSshDriver.java
ygy: Thanks for the comment.

Previously CouchDB was installed via binary packages (rpm, deb) that's why the binaries were located in /usr/bin, but the default configuration in the /etc. That's not more the case.
This PR does source installation for the CouchDB. It compiles in the `installDir` and than it copies the files in the `runDir`. `runDir` will be the prefix for all subdirs - bin, share, etc.

ygy:
@kiuby88 
There is a good explanation in our documentation - https://brooklyn.apache.org/v/latest/java/entities.html
It provides detailed information regarding the SoftwareProcess' Lifecycle. It contains a handful information regarding the drivers as well.

neykov:
Looks good, merging. One thing that could be surprising to users is that the `etc` folder gets overwritten on each start. Can be addressed if it becomes a problem though.

neykov:
Looks like when configuration is passed as a url it will not get used (see [here](https://github.com/ygy/brooklyn-library/blob/fix/couchdb/software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/couchdb/CouchDBNodeSshDriver.java#L194)). Can be addressed in a separate PR - not directly related to the changes in this one.

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/couchdb/CouchDBNodeSshDriver.java
kiuby88: Thanks for the explanation @ygy

neykov:
Ignore last comment - `COUCHDB_URI_TEMPLATE_URL` is not the URL alternative to `COUCHDB_CONFIG_TEMPLATE_URL` as I expected, but completely separate file.

PR #33, opened 2016-05-05
Allow creating roles for PostgreSQL

iyovcheva:


software/database/src/main/java/org/apache/brooklyn/entity/database/postgresql/PostgreSqlSshDriver.java
```diff
@@ -328,9 +340,39 @@ private void initializeNewDatabase() {
...
+
+        for (Map.Entry role: roles.entrySet()) {
+            createRolesQuery = createRolesQuery.concat(String.format("CREATE ROLE %s", role.getKey()));
+            if (((Map) role.getValue()).containsKey("properties")) {
+                createRolesQuery = createRolesQuery.concat(String.format(" WITH %s; ", ((Map) role.getValue()).get("properties")));
+            } else {
+                createRolesQuery = createRolesQuery.concat(";");
+            }
+
+            if (((Map) role.getValue()).containsKey("privilages")) {
```
geomacy: typo "privilages" -> "privileges"

software/database/src/main/java/org/apache/brooklyn/entity/database/postgresql/PostgreSqlNode.java
```diff
@@ -90,6 +95,29 @@
             String.class, "postgresql.username", "Username of the database user");
     
     String DEFAULT_USERNAME = "postgresqluser";
+
+    @SetFromFlag("roles")
+    ConfigKey<Map<String, Object>> ROLES = new MapConfigKey(Map.class, "postgresql.roles", "Initial roles with permissions",
+            ImmutableMap.<String, Object>of(
```
geomacy: I'd leave the default empty - hard to predict the roles users will want.

software/database/src/main/java/org/apache/brooklyn/entity/database/postgresql/PostgreSqlNode.java
```diff
@@ -90,6 +95,29 @@
...
+                            "privilages", ImmutableList.of("SELECT, INSERT, UPDATE, DELETE ON ALL TABLES IN SCHEMA public",
+                                    "EXECUTE ON ALL FUNCTIONS IN SCHEMA public")
+                    ),
+                    "Analyst", ImmutableMap.of(
+                            "privilages", "SELECT ON ALL TABLES IN SCHEMA public"
+                    )
+            ));
+
+    @SetFromFlag("createRoles")
+    ConfigKey<Boolean> CREATE_ROLES = ConfigKeys.newBooleanConfigKey("create.roles", "Should roles be created with database initialization", false);
```
geomacy: Do you really need this?   Why not just check if config key ROLES is not empty?

software/database/src/main/java/org/apache/brooklyn/entity/database/postgresql/PostgreSqlSshDriver.java
iyovcheva: Thanks @geomacy 
Do you think the roles defaults are fine to be in brooklyn as they look pretty standard, not specific?

iyovcheva:
Thanks @geomacy

software/database/src/main/java/org/apache/brooklyn/entity/database/postgresql/PostgreSqlNode.java
```diff
@@ -90,6 +95,17 @@
             String.class, "postgresql.username", "Username of the database user");
     
     String DEFAULT_USERNAME = "postgresqluser";
+
+    @SetFromFlag("roles")
+    ConfigKey<Map<String, Map>> ROLES = new MapConfigKey(Map.class, "postgresql.roles",
+            "Set roles with properties and permissions. Shoud be a map with keys equal to role names and values a map of the type:" +
+                  "key equal to `properties` and value - the role properties that should be in the query after `WITH` statement" +
+                    "key equal to `privileges` and value - the `GRANT` query value between the `GRANT` and `TO` statements"
+    );
```
tbouron: Maybe worth to add an actual example within the description of what the map should look like?

software/database/src/main/java/org/apache/brooklyn/entity/database/postgresql/PostgreSqlSshDriver.java
```diff
@@ -328,9 +340,39 @@ private void initializeNewDatabase() {
...
+            createRolesQuery = createRolesQuery.concat(String.format("CREATE ROLE %s", role.getKey()));
+            if (((Map) role.getValue()).containsKey(PostgreSqlNode.ROLE_PROPERTIES_KEY)) {
+                createRolesQuery = createRolesQuery.concat(String.format(" WITH %s; ", ((Map) role.getValue()).get("properties")));
+            } else {
+                createRolesQuery = createRolesQuery.concat(";");
+            }
+
+            if (((Map) role.getValue()).containsKey(PostgreSqlNode.ROLE_PRIVILEGES_KEY)) {
+                if ( ((Map) role.getValue()).get(PostgreSqlNode.ROLE_PRIVILEGES_KEY) instanceof List) {
+                    for (Object privilage: (List) ((Map) role.getValue()).get(PostgreSqlNode.ROLE_PRIVILEGES_KEY)) {
```
tbouron: s/privilage/privilege

software/database/src/main/java/org/apache/brooklyn/entity/database/postgresql/PostgreSqlSshDriver.java
```diff
@@ -328,9 +340,39 @@ private void initializeNewDatabase() {
...
+                        createRolesQuery = createRolesQuery.concat(String.format("GRANT %s TO %s; ", privilage, role.getKey()));
```
tbouron: s/privilage/privilege

software/database/src/main/java/org/apache/brooklyn/entity/database/postgresql/PostgreSqlSshDriver.java
```diff
@@ -328,9 +340,39 @@ private void initializeNewDatabase() {
...
                 callPgctl("stop", true))
                 .failOnNonZeroResultCode().execute();
     }
+
+    private String buildCreateRolesQuery() {
+        Map<String, Map> roles = entity.getConfig(PostgreSqlNode.ROLES);
+        String createRolesQuery = "\"";
+
+        for (Map.Entry role: roles.entrySet()) {
+            createRolesQuery = createRolesQuery.concat(String.format("CREATE ROLE %s", role.getKey()));
```
geomacy: Here and below, you shouldn't create SQL statements using string concatenation, that leaves you wide open to SQL injection attacks.  Instead use a prepared statement with appropriate bind variables, if at all possible.

tbouron: +1, always validate and escape input from the user, especially for SQL queries.

geomacy: Actually I'm not sure you're going to be able to use prepared statements with DDL like CREATE ROLE, at least not with Postgres.  (Worth a look to see whether it's supported anyway.)  At the very least, as Thomas says, you should be validating the input before using it in a statement.

duncangrant: I don't see the relevance.  This is a command populated from a blueprint where you can also add an database initialisation script that can do whatever you want.

tbouron: Ease of use I guess? Also, you could imagine to call this from an effector to create users and roles on the fly

duncangrant: Sorry - I meant I don't see the relevance of protecting against sql injection.

iyovcheva:
Thanks @geomacy @tbouron @duncangrant 
Using prepared statement requires big refactoring of the current code, because we generate and execute a script now and don't connect to the database in the code. Validating looks the simpler approach.

software/database/src/main/java/org/apache/brooklyn/entity/database/postgresql/PostgreSqlSshDriver.java
geomacy: @duncangrant you're right, if you can add arbitrary instructions in a script there's no point in worrying about injection  - my original comment didn't take that into account, it was just a gut reaction to seeing constructed SQL text.

nakomis:
I was a little late getting to this, but it looks like @tbouron, @geomacy, and @duncangrant  have already raised the concerns I had

Still not completely convinced that this is required in addition to the database initialisation script. Any thoughts anyone?

iyovcheva:
I agree, the user writing the queries in the yaml is the one who's deploying the entity and creating the database and can do whatever would like with them. But if this code is used for effectors in future or reused for anything, it's good to be secure from now.

software/database/src/main/java/org/apache/brooklyn/entity/database/postgresql/PostgreSqlSshDriver.java
```diff
@@ -328,9 +342,52 @@ private void initializeNewDatabase() {
...
                 sudoAsUser("postgres", getInstallDir() + "/bin/psql -p " + entity.getAttribute(PostgreSqlNode.POSTGRESQL_PORT) + 
                                 " --command="+ createDatabaseCommand),
+                createRolesAdditionalCommand,
                 callPgctl("stop", true))
                 .failOnNonZeroResultCode().execute();
     }
+
+    private String buildCreateRolesQuery() {
+        Map<String, Map> roles = entity.getConfig(PostgreSqlNode.ROLES);
+        String createRolesQuery = "\"";
```
bostko: It is better to use `StringBuilder`.

software/database/src/main/java/org/apache/brooklyn/entity/database/postgresql/PostgreSqlSshDriver.java
```diff
@@ -328,9 +342,52 @@ private void initializeNewDatabase() {
...
+
+        for (Map.Entry role: roles.entrySet()) {
+            createRolesQuery = createRolesQuery.concat(String.format("CREATE ROLE %s", validateInput((String) role.getKey())));
```
bostko: Most of the people use `+=` instead of `concat`

software/database/src/main/java/org/apache/brooklyn/entity/database/postgresql/PostgreSqlNode.java
```diff
@@ -90,6 +93,23 @@
             String.class, "postgresql.username", "Username of the database user");
     
     String DEFAULT_USERNAME = "postgresqluser";
+
+    @SetFromFlag("roles")
+    ConfigKey<Map<String, Map>> ROLES = new MapConfigKey(Map.class, "postgresql.roles",
```
aledsage: Can we have a test for this?

software/database/src/main/java/org/apache/brooklyn/entity/database/postgresql/PostgreSqlSshDriver.java
```diff
@@ -320,6 +324,16 @@ private void initializeNewDatabase() {
                 "\"CREATE DATABASE %s OWNER %s\"",
                 StringEscapes.escapeSql(getDatabaseName()),
                 StringEscapes.escapeSql(getUsername()));
+
+        String createRolesAdditionalCommand = "";
+
+        if (!entity.getConfig(PostgreSqlNode.ROLES).isEmpty()) {
```
aledsage: This looks like it will throw an NPE if `ROLES` config value has not been set (because it has no default value, such as an empty immutable map). Am I right?

software/database/src/main/java/org/apache/brooklyn/entity/database/postgresql/PostgreSqlSshDriver.java
```diff
@@ -328,9 +342,51 @@ private void initializeNewDatabase() {
...
+    private String validateInput(String input) {
+        String regex = "[A-Za-z_,\\s]+";
+        Pattern pattern = Pattern.compile(regex);
+        Matcher matcher = pattern.matcher(input);
+
+        if (matcher.find() && matcher.start() == 0 && matcher.end() == input.length()) {
+            return input;
+        }
+
+        return "";
```
aledsage: This seems like the wrong way to handle invalid input: we just return an empty string without logging any problem? And presumably the subsequent roles query would be invalid with missing role name etc, such as `CREATE ROLE  WITH ...`?

iyovcheva:
@aledsage are you ok with merging this?
I couldn't run the tests and after I resolve my issue with them, can make changes to `PostgreSqlInitializeDatabaseLiveTest` in another PR if needed.

aledsage:
@iyovcheva I've opened https://github.com/iyovcheva/brooklyn-library/pull/1 for some additional changes to the roles code. Can you take a look and see what you think please?

If you merge that, it should appear automatically on this PR.

iyovcheva:
retest this please

aledsage:
@iyovcheva there's a merge conflict - can you rebase against master please?

qa/src/test/resources/java-web-app-and-db-with-function.yaml
```diff
@@ -17,18 +17,20 @@
 # under the License.
 #
 name: java-cluster-db-example
-location: localhost
 services:
-- type: org.apache.brooklyn.entity.webapp.ControlledDynamicWebAppCluster
+- serviceType: org.apache.brooklyn.entity.webapp.ControlledDynamicWebAppCluster
```
aledsage: Why change this? `type` should work as well. This file seems unrelated to the rest of the pull request.

aledsage: The test `JavaWebAppsMatchingTest` works for me locally. Can you explain these changes please?

qa/src/test/resources/java-web-app-and-db-with-function.yaml
```diff
@@ -17,18 +17,20 @@
...
-  http.port: 9280+
-  proxy.http.port: 9210+
-  java.sysprops: 
-    brooklyn.example.db.url: $brooklyn:formatString("jdbc:%s%s?user=%s&password=%s",
-       component("db").attributeWhenReady("datastore.url"), "visitors", "brooklyn", "br00k11n")
-
-- type: org.apache.brooklyn.entity.database.mysql.MySqlNode
+  location: localhost
+  brooklyn.config:
+    wars.root: http://search.maven.org/remotecontent?filepath=io/brooklyn/example/brooklyn-example-hello-world-sql-webapp/0.7.0-M1/brooklyn-example-hello-world-sql-webapp-0.7.0-M1.war
```
aledsage: why change this url from the 0.9.0 to the 0.7.0-M1 version?!

qa/src/test/resources/java-web-app-and-db-with-function.yaml
```diff
@@ -21,12 +21,13 @@ location: localhost
 services:
 - type: org.apache.brooklyn.entity.webapp.ControlledDynamicWebAppCluster
   name: My Web
-  wars.root: http://repo1.maven.org/maven2/org/apache/brooklyn/example/brooklyn-example-hello-world-sql-webapp/0.9.0/brooklyn-example-hello-world-sql-webapp-0.9.0.war
-  http.port: 9280+
-  proxy.http.port: 9210+
-  java.sysprops: 
-    brooklyn.example.db.url: $brooklyn:formatString("jdbc:%s%s?user=%s&password=%s",
-       component("db").attributeWhenReady("datastore.url"), "visitors", "brooklyn", "br00k11n")
+  brooklyn.config:
```
aledsage: I'm surprised this is needed, but it's most probably a good thing to include the `brooklyn.config`, so happy to merge.

aledsage:
Thanks @iyovcheva - merging now.

software/database/pom.xml
```diff
@@ -144,11 +144,24 @@
...
+        <dependency>
             <groupId>postgresql</groupId>
             <artifactId>postgresql</artifactId>
             <version>${postgresql.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>javax.ws.rs</groupId>
+            <artifactId>javax.ws.rs-api</artifactId>
+            <version>2.0.1</version>
```
aledsage: Why do we need to include `javax.ws.rs-api` now? Also, it should use `${jax-rs-api.version}` rather than hard-coding 2.0.1.

software/database/pom.xml
```diff
@@ -144,11 +144,24 @@
         </dependency>
 
         <dependency>
+            <groupId>${jclouds.groupId}.provider</groupId>
```
aledsage: We already depend on `brooklyn-locations-jclouds`, which should bring in this dependency. Why did it need to be added?

neykov: Maven test dependencies are not transitive. Need to be explicit about test dependencies.

aledsage: Confirmed that we can delete the aws-ec2 dependency, but we do need to keep the `javax.ws.rs-api` dependency for the live tests to work (in eclipse at least). We should change it to use `${jax-rs-api.version}` though.

PR #34, opened 2016-05-09
Fix sed command that update the MySQL configuration file upon password change

tbouron:


software/database/src/main/java/org/apache/brooklyn/entity/database/mysql/MySqlNodeEffectors.java
```diff
@@ -75,7 +75,7 @@ public Void call(ConfigBag parameters) {
             DynamicTasks.queue(
                     SshTasks.newSshExecTaskFactory(machine, 
                                     "cd "+entity().getAttribute(MySqlNode.RUN_DIR),
-                                    "sed -i'' -e 's@^\\(\\s*password\\s*=\\s*\\).*$@\\1" + newPass.replace("\\", "\\\\") + "@g' mymysql.cnf")
+                                    "sed -i.bak 's/\\(password.*\\)=.*/\\1= " + newPass.replace("\\", "\\\\") + "/' mymysql.cnf")
```
neykov: @tbouron Won't this replace lines containing the following as well?
```
passwordMine=xxx
passwordOld=yyy
passwordDontChange=zzz
```

What about `password=se!=scret`. I think it will match greedily here, so won't do the replacement right?

Is the change fixing the case for passwords containing `@`? Could you add some more details in the commit message.

aledsage:
@tbouron what do you think of @neykov 's comments?

What problem do you hit with the old code, and on what environment(s)?

I believe the old code is trying to support OSX and Linux (but probably doesn't manage to). That's tricky because `sed -i` behaves differently (see discussion in http://stackoverflow.com/questions/2320564/variations-of-sed-between-osx-and-gnu-linux).

Can we add an integration/live test for this?

As an aside: we can almost test this in isolation as an integration test, because `ChangePasswordEffectorBody` is defined separately from the `MySqlNode`! Except for the fact that the impl does `((MySqlNodeImpl)entity()).getDriver()`. If it wasn't for that, you could create a `VanillaSoftwareProcess`, wire in that effector, create a `mymysql.cnf` file in its run dir, set the attribute `MySqlNode.PASSWORD` on your test entity, and then call the effector. But probably not worth trying to refactor to make that possible.

software/database/src/main/java/org/apache/brooklyn/entity/database/mysql/MySqlNodeEffectors.java
tbouron: You are right for the 2 mentioned cases @neykov, I didn't think this through.

I'll need to go back to the drawing board.

tbouron:
@aledsage I encountered issues in all environments, any cloud. Even though the password is changed at the MySQL level, it never updates the `mymysql.cnf` file, probably because of absence of the `-i` option. As a result, the entity goes on fire because the `checkRunning` command uses the credentials specified within this file.

We have a "[live test](https://github.com/brooklyncentral/brooklyn-mysql-node/blob/master/catalog.tests.bom)", that's how I noticed this issue.

tbouron:
Close in favor or #76 (my PR wasn't really working anyway)

PR #35, opened 2016-05-10
New EtcdCluster entities

grkvlt:
As used in [brooklyncentral/clocker](https://github.com/brooklyncentral/clocker)


grkvlt:
Moving to external repository instead of core Brooklyn code

PR #37, opened 2016-05-18
Avoid deadlock in AbstractGeoDnsServiceImpl

sjcorbett:
See commit message in 2e1246a for details.


sjcorbett:
I also need to change `GeoscalingDnsServiceImpl` to not synchronise on the instance.

aledsage:
LGTM. Merging.

When hitting something serious such as a deadlock, I like to include a copy of the stacktraces in a jira issue (or at least in the comment of the pull request). That means it will be indexed by google, so anyone else who hits it with an older Brooklyn version can find details. But no need to do anything extra for this PR.

PR #38, opened 2016-05-20
Fix BROOKLYN-271: Avoiding object hashcode in entities' toString

kiuby88:
Fixing [BROOKLYN-271](https://issues.apache.org/jira/browse/BROOKLYN-271)
Avoiding object references in catalog ConfigKey cards. Then, `toString` method of `NginxDefaultConfigGenerator` has to return just the class name.


kiuby88:
@aledsage this PR is involved with https://github.com/apache/brooklyn-server/pull/146

aledsage:
LGTM - merging. Thanks @kiuby88

PR #44, opened 2016-06-17
RiakNode Internal Networking

grkvlt:
Configure the internal ports required for Riak intra-node communication using an optional `JcloudsSecurityGroupCustomizer`

### DIFF
--- a/software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/riak/RiakNode.java
+++ b/software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/riak/RiakNode.java
@@ -23,6 +23,7 @@ import java.util.List;
 
 import org.apache.brooklyn.api.catalog.Catalog;
 import org.apache.brooklyn.api.entity.ImplementedBy;
+import org.apache.brooklyn.api.location.PortRange;
 import org.apache.brooklyn.api.sensor.AttributeSensor;
 import org.apache.brooklyn.config.ConfigKey;
 import org.apache.brooklyn.core.annotation.Effector;
@@ -30,6 +31,7 @@ import org.apache.brooklyn.core.annotation.EffectorParam;
 import org.apache.brooklyn.core.config.ConfigKeys;
 import org.apache.brooklyn.core.effector.MethodEffector;
 import org.apache.brooklyn.core.entity.Attributes;
+import org.apache.brooklyn.core.location.PortRanges;
 import org.apache.brooklyn.core.sensor.AttributeSensorAndConfigKey;
 import org.apache.brooklyn.core.sensor.PortAttributeSensorAndConfigKey;
 import org.apache.brooklyn.core.sensor.Sensors;
@@ -107,10 +109,10 @@ public interface RiakNode extends SoftwareProcess, UsesJava {
 
     // NB these two needed for clients to access
     @SetFromFlag("riakWebPort")
-    PortAttributeSensorAndConfigKey RIAK_WEB_PORT = new PortAttributeSensorAndConfigKey("riak.webPort", "Riak Web Port", "8098+");
+    PortAttributeSensorAndConfigKey RIAK_WEB_PORT = ConfigKeys.newPortSensorAndConfigKey("riak.web.port", "Riak Web Port", "8098+");
 
     @SetFromFlag("riakPbPort")
-    PortAttributeSensorAndConfigKey RIAK_PB_PORT = new PortAttributeSensorAndConfigKey("riak.pbPort", "Riak Protocol Buffers Port", "8087+");
+    PortAttributeSensorAndConfigKey RIAK_PB_PORT = ConfigKeys.newPortSensorAndConfigKey("riak.pb.port", "Riak Protocol Buffers Port", "8087+");
 
     @SetFromFlag("useHttpMonitoring")
     ConfigKey<Boolean> USE_HTTP_MONITORING = ConfigKeys.newConfigKey("httpMonitoring.enabled", "HTTP(S) monitoring enabled", Boolean.TRUE);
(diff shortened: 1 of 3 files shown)


software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/riak/RiakNode.java
```diff
@@ -125,14 +127,37 @@
...
-    PortAttributeSensorAndConfigKey EPMD_LISTENER_PORT = new PortAttributeSensorAndConfigKey("epmdListenerPort", "Erlang Port Mapper Daemon Listener Port", "4369");
-    PortAttributeSensorAndConfigKey ERLANG_PORT_RANGE_START = new PortAttributeSensorAndConfigKey("erlangPortRangeStart", "Erlang Port Range Start", "6000+");
-    PortAttributeSensorAndConfigKey ERLANG_PORT_RANGE_END = new PortAttributeSensorAndConfigKey("erlangPortRangeEnd", "Erlang Port Range End", "7999+");
+    /*
+     * Needed for nodes to talk to each other, but not clients, so ideally set up in the security group for internal access and configured here.
+     * If {@link #CONFIGURE_INTERNAL_NETWORKING} is set, then a location customizer will be added to confiure the security group dynamically.
+     */
+
+    @SetFromFlag("handoffListenerPort")
+    ConfigKey<Integer> HANDOFF_LISTENER_PORT = ConfigKeys.newIntegerConfigKey("riak.handoff.port.internal", "Handoff Listener Port", 8099);
```
geomacy: should the key be named '....internal.port'   i.e. end in ".port", in order to be recognised as a port configuration parameter?  And same below.

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/riak/RiakNodeImpl.java
```diff
@@ -99,17 +110,59 @@ public boolean isPackageDownloadUrlProvided() {
...
+        Location location = getDriver().getLocation();
+        if (!(location instanceof JcloudsSshMachineLocation)) {
+            LOG.info("Not running in a JcloudsSshMachineLocation, not adding IP permissions to {}", this);
+            return;
+        }
+        JcloudsMachineLocation machine = (JcloudsMachineLocation) location;
+        JcloudsLocationSecurityGroupCustomizer customizer = JcloudsLocationSecurityGroupCustomizer.getInstance(getApplicationId());
+
+        synchronized (getParent()) {
+            String cidr = Cidr.UNIVERSAL.toString(); // TODO use more appropriate CIDR
```
geomacy: maybe update comment to explain what would be more appropriate (what's wrong with this line as it stands?)

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/riak/RiakNode.java
grkvlt: No, because they're not proper port config like you normall have named like that. The `.internal` suffix may change, depending on how networking extensions are implemented in future, though.

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/riak/RiakNodeImpl.java
grkvlt: Yes, I mean more _restrictive_ CIDR, really. But I don't think we have a good way of getting the network address for the hosts in Brooklyn at present.

geomacy:
Looks good to me, but haven't tested it out myself.

grkvlt:
@geomacy I'm going to try this out with the latest suite of Brooklyn/AMP/Clocker/jclouds-labs patches, and will report back

grkvlt:
@geomacy Tested this, and it works fine.

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/riak/RiakNode.java
aledsage: Let's indicate that it's `@beta` (e.g. in the description, and as a java annotation) if we think that the name will change?

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/riak/RiakNodeImpl.java
```diff
@@ -99,17 +110,59 @@ public boolean isPackageDownloadUrlProvided() {
...
     protected Collection<Integer> getRequiredOpenPorts() {
-        // TODO this creates a huge list of inbound ports; much better to define on a security group using range syntax!
-        int erlangRangeStart = getConfig(ERLANG_PORT_RANGE_START).iterator().next();
-        int erlangRangeEnd = getConfig(ERLANG_PORT_RANGE_END).iterator().next();
+        PortRange erlangPorts = config().get(ERLANG_PORT_RANGE);
+        Integer erlangRangeStart = config().get(ERLANG_PORT_RANGE_START);
+        Integer erlangRangeEnd = config().get(ERLANG_PORT_RANGE_END);
+        if (erlangRangeStart == null) erlangRangeStart = Iterables.get(erlangPorts, 0);
+        if (erlangRangeEnd == null) erlangRangeEnd = Iterables.getLast(erlangPorts);
+        sensors().set(ERLANG_PORT_RANGE_START, erlangRangeStart);
```
aledsage: I'm a bit confused by the deprecation of this. I'd have expected:
* a `log.warn` when using the deprecated value (i.e. when it's not null), to alert people that they are using the deprecated value
* us not to use the sensor (other than to set it for backwards compatibility), given it's deprecated - but it's later used in `configureInternalNetwork`.
* changing the key's name might cause backwards compatibility problems (even though you have `@SetFromFlag` with the old name). The setFromFlag stuff doesn't work for inherited config, I believe.
* I'd have expected you to keep the old key (deprecated) and declare the new key; to read the value from both and to warn if `getRaw()` is non-null for the old one. If you do that, then we are free to delete it at some point. The way you've changed it, then it's hard to ever delete the `@SetFromFlag` name.

---
Using PortRange doesn't feel quite right. Our `PortRange` does not have to be linear (e.g. you could do `PortRanges.fromString("1,3,5")`).

We subsequently use it as though it must be a contiguous range, with a simple start and end value. Maybe we should just leave the configuration as that?

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/riak/RiakNodeImpl.java
```diff
@@ -99,17 +110,59 @@ public boolean isPackageDownloadUrlProvided() {
...
+    private void configureInternalNetworking() {
+        Location location = getDriver().getLocation();
+        if (!(location instanceof JcloudsSshMachineLocation)) {
+            LOG.info("Not running in a JcloudsSshMachineLocation, not adding IP permissions to {}", this);
+            return;
+        }
+        JcloudsMachineLocation machine = (JcloudsMachineLocation) location;
+        JcloudsLocationSecurityGroupCustomizer customizer = JcloudsLocationSecurityGroupCustomizer.getInstance(getApplicationId());
+
+        synchronized (getParent()) {
```
aledsage: Why synchronize on the parent? That feels like bad practice. Is there something better to synchronize on? Or can the `JcloudsLocationSecurityGroupCustomizer` handle the synchronizing internally?

grkvlt: Agree there is possibly something better that we could choose. I'll have another look, when I'm updating the code.

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/riak/RiakNodeImpl.java
grkvlt: @aledsage I'll revisit this - perhaps better to just have two integer config keys/sensors, and no port range, and avoid any deprecation.

grkvlt:
@aledsage you were right, the customizer synchronises on its class, so no need for external locking. I also removed the deprecation message, but kept the new config and sensor names, as I believe this is a safe change to make.

grkvlt:
@aledsage I'm going to merge this if there are no further objections

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/riak/RiakNode.java
```diff
@@ -107,10 +107,10 @@
 
     // NB these two needed for clients to access
     @SetFromFlag("riakWebPort")
-    PortAttributeSensorAndConfigKey RIAK_WEB_PORT = new PortAttributeSensorAndConfigKey("riak.webPort", "Riak Web Port", "8098+");
+    PortAttributeSensorAndConfigKey RIAK_WEB_PORT = ConfigKeys.newPortSensorAndConfigKey("riak.web.port", "Riak Web Port", "8098+");
```
aledsage: Changing this name breaks backwards compatibility. If someone has in their blueprint `riak.webPort: 1234` then that config option will now be ignored. Same for `riak.pbPort: 1235`.

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/riak/RiakNode.java
```diff
@@ -125,14 +125,28 @@
...
-    PortAttributeSensorAndConfigKey EPMD_LISTENER_PORT = new PortAttributeSensorAndConfigKey("epmdListenerPort", "Erlang Port Mapper Daemon Listener Port", "4369");
-    PortAttributeSensorAndConfigKey ERLANG_PORT_RANGE_START = new PortAttributeSensorAndConfigKey("erlangPortRangeStart", "Erlang Port Range Start", "6000+");
-    PortAttributeSensorAndConfigKey ERLANG_PORT_RANGE_END = new PortAttributeSensorAndConfigKey("erlangPortRangeEnd", "Erlang Port Range End", "7999+");
+    /*
+     * Needed for nodes to talk to each other, but not clients, so ideally set up in the security group for internal access and configured here.
+     * If {@link #CONFIGURE_INTERNAL_NETWORKING} is set, then a location customizer will be added to confiure the security group dynamically.
+     */
+
+    @SetFromFlag("handoffListenerPort")
+    ConfigKey<Integer> HANDOFF_LISTENER_PORT = ConfigKeys.newIntegerConfigKey("riak.handoff.port", "Handoff Listener Port", 8099);
```
aledsage: This is no longer published as a sensor, which changes backwards compatibility. It probably won't affect anyone, but you can instead use:

    @SetFromFlag("handoffListenerPort")
    AttributeSensorAndConfigKey<Integer, Integer> HANDOFF_LISTENER_PORT = ConfigKeys.newSensorAndConfigKey(
            Integer.class,
            "riak.handoff.port",
            "Handoff Listener Port", 
            8099);

Probably worth it, just in case.

neykov: This still won't publish the sensor. Only `PortRange`s get published automatically.

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/riak/RiakNode.java
aledsage: I'd be very cautious: I'd create a second config key with the old name, and fallback to that. And I'd make sure a sensor with the old name is published.

Looking at some customer bespoke code, they depend on "riak.pbPort" and "riak.webPort" being published as sensors (for their enricher that propagates the public endpoint). We can tell them to change this, but we should not break backwards compatibility lightly.

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/riak/RiakNode.java
aledsage: @neykov are you sure? I thought `ConfigToAttributes.apply(Entity)` would get called, which would do this for anything of type `AttributeSensorAndConfigKey`. But I haven't tested this recently.

aledsage:
@grkvlt finished reviewing again. Can you have a look at the backwards compatibility issue I've highlighted please.

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/riak/RiakNode.java
neykov: That's right - checked the sources. Please ignore.

grkvlt:
@aledsage I made the compatibility changes you suggested

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/riak/RiakNodeImpl.java
```diff
@@ -99,15 +108,51 @@ public boolean isPackageDownloadUrlProvided() {
 
     @Override
     protected Collection<Integer> getRequiredOpenPorts() {
-        // TODO this creates a huge list of inbound ports; much better to define on a security group using range syntax!
-        int erlangRangeStart = getConfig(ERLANG_PORT_RANGE_START).iterator().next();
-        int erlangRangeEnd = getConfig(ERLANG_PORT_RANGE_END).iterator().next();
+        Integer erlangRangeStart = config().get(ERLANG_PORT_RANGE_START);
+        Integer erlangRangeEnd = config().get(ERLANG_PORT_RANGE_END);
+        sensors().set(ERLANG_PORT_RANGE_START, erlangRangeStart);
```
aledsage: I suspect we get this being set for free, even if you don't include this line - but no strong feelings about leaving it in.

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/riak/RiakNodeImpl.java
```diff
@@ -99,15 +108,51 @@ public boolean isPackageDownloadUrlProvided() {
...
+        sensors().set(ERLANG_PORT_RANGE_END, erlangRangeEnd);
+
+        boolean configureInternalNetworking = config().get(CONFIGURE_INTERNAL_NETWORKING);
```
aledsage: I'd do the `Boolean.TRUE.equals(config().get(CONFIGURE_INTERNAL_NETWORKING))`, in case of null.

aledsage:
Thanks @grkvlt - I'm going to merge now. I made one other minor comment that is perhaps worth addressing, but not worth holding up this PR for.

PR #47, opened 2016-07-04
Use Entities.descendantsAndSelf

aledsage:
Previously calling Entities.deprecated; that is not deprecated.

This is dependent on first merging https://github.com/apache/brooklyn-server/pull/218


neykov:
LGTM, merging.

PR #48, opened 2016-07-07
Fix NodeJsWebAppSshDriver to obtain properly MAIN_URL

bostko:
- make NodeJsWebAppSshDriver and JavaWebAppSshDriver to use a single method
  for obtaining the MAIN_URL

This should be a standardized for all entities which has to have
an option for exposing http MAIN_URL

### DIFF
--- a/software/webapp/src/main/java/org/apache/brooklyn/entity/webapp/JavaWebAppSshDriver.java
+++ b/software/webapp/src/main/java/org/apache/brooklyn/entity/webapp/JavaWebAppSshDriver.java
@@ -91,25 +91,17 @@ public abstract class JavaWebAppSshDriver extends JavaSoftwareProcessSshDriver i
         return (ssl == null) ? null : ssl.getKeyAlias();
     }
 
+    /**
+     * @deprecated since 0.10.0; please use {@link WebAppServiceMethods#inferBrooklynAccessibleRootUrl(org.apache.brooklyn.api.entity.Entity)}
+     */
+    @Deprecated
     protected String inferRootUrl() {
-        if (isProtocolEnabled("https")) {
-            Integer port = getHttpsPort();
-            checkNotNull(port, "HTTPS_PORT sensors not set; is an acceptable port available?");
-            HostAndPort accessibleAddress = BrooklynAccessUtils.getBrooklynAccessibleAddress(getEntity(), port);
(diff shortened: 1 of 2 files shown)


software/webapp/src/main/java/org/apache/brooklyn/entity/webapp/WebAppServiceMethods.java
```diff
@@ -73,17 +73,26 @@ public static boolean isProtocolEnabled(Entity entity, String protocol) {
...
-            checkNotNull(rawPort, "HTTP_PORT sensors not set for %s; is an acceptable port available?", entity);
-            HostAndPort hp = BrooklynAccessUtils.getBrooklynAccessibleAddress(entity, rawPort);
-            return String.format("http://%s:%s/", hp.getHostText(), hp.getPort());
+            return inferBrooklynAccessibleRootUrl(entity, HTTP_PORT);
         } else {
             throw new IllegalStateException("HTTP and HTTPS protocols not enabled for "+entity+"; enabled protocols are "+getEnabledProtocols(entity));
         }
     }
+
+    public static String inferBrooklynAccessibleRootUrl(Entity entity, AttributeSensor<Integer> sensor) {
```
iyovcheva: I don't think we need inferBrooklynAccessibleRootUrl(Entity entity, AttributeSensor<Integer> sensor) signature as there are not defined other protocols yet and there is double check once in inferBrooklynAccessibleRootUrl(Entity entity) and once here. A sensor can be set in the first method instead and this would simplify the code.

iyovcheva:
LGTM, minor comment only

software/webapp/src/main/java/org/apache/brooklyn/entity/webapp/JavaWebAppSshDriver.java
```diff
@@ -91,25 +91,17 @@ protected String getSslKeyAlias() {
         return (ssl == null) ? null : ssl.getKeyAlias();
     }
 
+    /**
+     * @deprecated please use {@link WebAppServiceMethods#inferBrooklynAccessibleRootUrl(org.apache.brooklyn.api.entity.Entity)}
```
aledsage: It's a good idea to include `@deprecated since 0.10.0; ...`.

aledsage:
LGTM; one tiny comment and then good to merge.

neykov:
Thanks all, merging.

PR #50, opened 2016-07-13
Adds missing logos for library entities

aledsage:


aledsage:
Reviewed by @Graeme-Miller - merging now.

PR #52, opened 2016-07-20
Fix riak (configureInternalNetwork throwing NPE)

aledsage:
Calling `configureInternalNetworking()` in `getRequiredOpenPorts()` always caused an NPE, because the driver was not yet initialised. Also the sensor `ERLANG_PORT_RANGE_START` was always null, because the config had no default value.

We unfortunately broke this in https://github.com/apache/brooklyn-library/pull/44.


iyovcheva:
Looks good and testing passed

aledsage:
Thanks @iyovcheva - merging now.

PR #53, opened 2016-07-22
Move karaf distribution to brooklyn-dist.

geomacy:
This PR is the second of three that together move the karaf distribution to brooklyn-dist and separate out the software from brooklyn-library into separate catalog bom and feature files.

This change separates out the brooklyn-library releated features into
new projects in brooklyn-library itself.

Merge this after https://github.com/apache/brooklyn-server/pull/271.

### DIFF
--- a/karaf/catalog/pom.xml
+++ b/karaf/catalog/pom.xml
@@ -1,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
+    distributed with this work for additional information
+    regarding copyright ownership.  The ASF licenses this file
+    to you under the Apache License, Version 2.0 (the
+    "License"); you may not use this file except in compliance
+    with the License.  You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing,
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.apache.brooklyn</groupId>
+        <artifactId>brooklyn-library-karaf</artifactId>
+        <version>0.10.0-SNAPSHOT</version>  <!-- BROOKLYN_VERSION -->
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+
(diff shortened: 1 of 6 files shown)


karaf/features/src/main/feature/feature.xml
```diff
@@ -0,0 +1,97 @@
...
+
+    <feature name="brooklyn-software-monitoring" version="${project.version}" description="Brooklyn Monitoring Software Entities">
+        <bundle>mvn:org.apache.brooklyn/brooklyn-software-monitoring/${project.version}</bundle>
+        <feature>brooklyn-core</feature>
+        <feature>brooklyn-api</feature>
+        <feature>brooklyn-software-base</feature>
+        <bundle dependency="true">mvn:com.google.guava/guava/${guava.version}</bundle>
+    </feature>
+
+    <!-- TODO Don't load all bundles out of the box in the long run. Let users cherry pick what they need. -->
```
neykov: I think this comment can be removed - bundling libraries is what we want now.

karaf/features/src/main/feature/feature.xml
```diff
@@ -0,0 +1,97 @@
...
+          name="org.apache.brooklyn-${project.version}"
+          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:noNamespaceSchemaLocation="http://karaf.apache.org/xmlns/features/v1.2.0">
+
+    <repository>mvn:org.apache.karaf.features/standard/${karaf.version}/xml/features</repository>
+    <repository>mvn:org.apache.karaf.features/enterprise/${karaf.version}/xml/features</repository>
+    <repository>mvn:org.apache.karaf.features/spring/${karaf.version}/xml/features</repository>
+
+    <repository>mvn:org.apache.jclouds.karaf/jclouds-karaf/${jclouds.version}/xml/features</repository>
+    <repository>mvn:org.apache.cxf.karaf/apache-cxf/${cxf.version}/xml/features</repository>
```
neykov: Don't think we need any of these repos here?

karaf/features/pom.xml
```diff
@@ -0,0 +1,60 @@
...
+                </plugin>
+            </plugins>
+        </pluginManagement>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.karaf.tooling</groupId>
+                <artifactId>karaf-maven-plugin</artifactId>
+
+                <configuration>
+                    <startLevel>50</startLevel>
```
neykov: Why override default startLevel?

karaf/catalog/pom.xml
```diff
@@ -0,0 +1,54 @@
...
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <extensions>true</extensions>
+                <configuration>
+                    <instructions>
+                        <Import-Package>*</Import-Package>
```
neykov: Isn't that the default? Can remove the config.

karaf/features/pom.xml
geomacy: It's a good question - I copied this as-is from brooklyn-server/karaf/features, where it was introduced by https://github.com/apache/brooklyn-server/commit/4b1ede22489e18605829d91b4af6eb2a0c41645d, but it isn't obvious to me that this would be necessary.

geomacy:
Review comments applied, squashed and pushed.

duncangrant:
I've tested this - all seems to work

PR #55, opened 2016-08-02
AbstractControllerImpl: expose http url by default

bostko:


software/webapp/src/main/java/org/apache/brooklyn/entity/proxy/AbstractControllerImpl.java
```diff
@@ -287,7 +287,7 @@ protected String inferUrl(boolean requireManagementAccessible) {
     }
 
     protected String inferUrl() {
-        return inferUrl(false);
+        return inferUrl(true);
```
aledsage: This doesn't feel right - it will mean that "main.uri" will have the publicly mapped address (e.g. if running in vCloud Director behind a DNAT rule then it will show the dnat'ed port).

That contradicts the proposal in dev@brooklyn (email subject "Working with Multiple Networks - Apache Brooklyn Feature Proposal", link to proposal is 
https://docs.google.com/document/d/1IrWLWunWSl_ScwY3MRICped8eJMjQEH1FbWZJcoK0Iw). That proposal says the main.uri should be the subnet address.

neykov:
bump @bostko
Is this still an issue? Can you answer Aled's question from above.

bostko:
This PR should be revisited. Closing for now.

PR #56, opened 2016-08-03
Update Nginx config file generator and add support for connection upgrade

grkvlt:
Also fixes an NPE when no SSL config is provided


aledsage:
@grkvlt what tests should we run to confirm that this hasn't broken previous use-cases, and that it  now supports the use-case that motivated you to make this change?

grkvlt:
@aledsage the existing Nginx tests are sufficient to verify there are no regressions, however the use-case for this code is no longer as urgent. It is still a useful change, since it supports HTTP 1.1 and connection upgrades properly now, and the code is also more readable.

The change to the config file that adds this is the addition of these two groups of lines:

```
    map $http_upgrade $connection_upgrade {
      default Upgrade;
      ''      close;
    }
```
```
    proxy_http_version 1.1;
    proxy_set_header Upgrade $http_upgrade;
    proxy_set_header Connection $connection_upgrade;
```

neykov:
@grkvlt the `server.conf` used with `NginxTemplateConfigGenerator` follows roughly the config from `NginxDefaultConfigGenerator`. Can you update it as well.
I think we should just migrate to the template configuration and deprecate `NginxDefaultConfigGenerator` altogether as its easier to support.

nakomis:
@grkvlt Can you take a look at this please to see if it is still relevant, and address the comments above if appropriate

Thanks

nakomis:
P.S. Can you also rebase if this is still valid

PR #58, opened 2016-08-08
Added perl install to MySQL node

drigodwin:
Addresses [BROOKLYN-300](https://issues.apache.org/jira/browse/BROOKLYN-300).

Tested with Centos 6 and Centos 7 on BlueBox


aledsage:
LGTM; merging now.

PR #59, opened 2016-08-15
Fix NPE generating Nginx config when no SSL configured

grkvlt:
Extracts NPE fix from #56


grkvlt:
Simple change to fix obvious bug, so merging as-is

geomacy:
+1 LGTM and fixes the NPE I was seeing when deploying deploy-multi-node-app-swarm while testing https://github.com/cloudsoft/container-service/pull/60

PR #62, opened 2016-08-25
BIND DNS improvements

ahgittin:
support configurable source sensor for the IP address (so we can do private or public), plus some misc fixes for greater OS portability


software/network/src/main/resources/org/apache/brooklyn/entity/network/bind/named.conf
```diff
@@ -20,14 +20,15 @@
 // Generated by Brooklyn on ${.now?string.short}
 //
 options {
-    listen-on port ${entity.dnsPort} { 127.0.0.1; ${driver.address}; };
+    // listen-on default is all interfaces
+    // listen-on port ${entity.dnsPort} { 127.0.0.1; ${driver.address}; ${driver.subnetAddress}; };
```
grkvlt: would prefer `listen-on port ${entity.dnsPort} { any; };` to be explicit, and allow changing port number, then. same for IPV6 as `listen-on-v6 port ${entity.dnsPort} { any; };`

software/network/src/main/java/org/apache/brooklyn/entity/network/bind/BindDnsServerImpl.java
```diff
@@ -222,17 +227,35 @@ public void update() {
...
+            for (Map.Entry<String, Entity> entry : hostnameToEntity.entries()) {
+                String domainName = entry.getKey();
+                Entity entity = entry.getValue();
+                
+                String address = null;
+                
+                AttributeSensor<String> addressSensor = getConfig(ADDRESS_SENSOR);
+                if (addressSensor!=null) {
+                    address = entity.getAttribute(addressSensor);
+
```
grkvlt: Remove blank line

grkvlt:
LGTM, can be merged once the `listen-on port` lines are fixed

software/network/src/main/resources/org/apache/brooklyn/entity/network/bind/named.conf
grkvlt: Actually, use the following for both:
```
// This will cause named to listen on all IPv4 and IPv6 configured interfaces.
listen-on-v6 ${entity.dnsPort} { any; };
```

ahgittin:
good suggestions, but as discussed given the testing time required, low priority of ipv6, and reality that we should rewrite much of this in yaml, i've marked the (very nice) ipv6 change as `TODO` in the conf file

PR #64, opened 2016-09-26
BROOKLYN-350 Improving the download of external packages

ygy:
- externalizing the CURL command
- adding default parameter so curl follows location redirects

### DIFF
--- a/software/database/src/main/java/org/apache/brooklyn/entity/database/postgresql/PostgreSqlNode.java
+++ b/software/database/src/main/java/org/apache/brooklyn/entity/database/postgresql/PostgreSqlNode.java
@@ -52,7 +52,7 @@ import java.util.Map;
 public interface PostgreSqlNode extends SoftwareProcess, HasShortName, DatastoreCommon, DatabaseNode {
     
     @SetFromFlag("version")
-    ConfigKey<String> SUGGESTED_VERSION = ConfigKeys.newConfigKeyWithDefault(SoftwareProcess.SUGGESTED_VERSION, "9.3-1");//"9.1-4");
+    ConfigKey<String> SUGGESTED_VERSION = ConfigKeys.newConfigKeyWithDefault(SoftwareProcess.SUGGESTED_VERSION, "9.3-3");//"9.1-4");
 
     @SetFromFlag("configFileUrl")
     ConfigKey<String> CONFIGURATION_FILE_URL = ConfigKeys.newStringConfigKey(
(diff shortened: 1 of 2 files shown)


software/database/src/main/java/org/apache/brooklyn/entity/database/postgresql/PostgreSqlNode.java
```diff
@@ -52,13 +52,17 @@
...
+    ConfigKey<String> SUGGESTED_VERSION = ConfigKeys.newConfigKeyWithDefault(SoftwareProcess.SUGGESTED_VERSION, "9.3-3");//"9.1-4");
 
     @SetFromFlag("configFileUrl")
     ConfigKey<String> CONFIGURATION_FILE_URL = ConfigKeys.newStringConfigKey(
             "postgresql.config.file.url", "URL where PostgreSQL configuration file can be found; "
                 + "if not supplied the blueprint uses the default and customises it");
 
+    @SetFromFlag("curlCommand")
+    ConfigKey<String> CURL_COMMAND = ConfigKeys.newStringConfigKey(
+            "curl.command", "CURL comamnd with parameters to be used", "curl -L");
```
bostko: Could you rename it to `cURL`.
Could you try adding more info on this parameter?

bostko:
@ygy Not very intuitive approach with using a ConfigKey. Still could be sensible in some situations.

software/database/src/main/java/org/apache/brooklyn/entity/database/postgresql/PostgreSqlNode.java
```diff
@@ -52,13 +52,17 @@
...
+    ConfigKey<String> SUGGESTED_VERSION = ConfigKeys.newConfigKeyWithDefault(SoftwareProcess.SUGGESTED_VERSION, "9.3-3");//"9.1-4");
 
     @SetFromFlag("configFileUrl")
     ConfigKey<String> CONFIGURATION_FILE_URL = ConfigKeys.newStringConfigKey(
             "postgresql.config.file.url", "URL where PostgreSQL configuration file can be found; "
                 + "if not supplied the blueprint uses the default and customises it");
 
+    @SetFromFlag("curlCommand")
+    ConfigKey<String> CURL_COMMAND = ConfigKeys.newStringConfigKey(
+            "curl.command", "cURL comamnd to be used. It follows location redirects by default (added -L parameter)", "curl -L");
```
bostko: Strange config usage of а parameter.
Very important to add a sentence about checking the parameter usage in Java code.

I would rather declare a CURL_COMMAND_OPTIONS parameter and put the single `-L` as a default value.

neykov: +1 could even do away with the config altogether - that's not something I've ever seen needing configuration. No strong opinion though.

ygy:
@bostko @neykov Can you review the new version?

bostko:
@ygy Looks good!

neykov:
+1, merging.

PR #66, opened 2016-10-11
Add more sensors for main.uri, based on the context, i.e public or not

tbouron:
Create 2 new sensors: `main.uri.mapped.public` and `main.uri.mapped.subnet`

Requires https://github.com/apache/brooklyn-server/pull/378

### DIFF
--- a/software/webapp/src/main/java/org/apache/brooklyn/entity/proxy/AbstractControllerImpl.java
+++ b/software/webapp/src/main/java/org/apache/brooklyn/entity/proxy/AbstractControllerImpl.java
@@ -33,6 +33,7 @@ import org.apache.brooklyn.api.mgmt.Task;
 import org.apache.brooklyn.api.policy.Policy;
 import org.apache.brooklyn.api.policy.PolicySpec;
 import org.apache.brooklyn.api.sensor.AttributeSensor;
+import org.apache.brooklyn.core.entity.Attributes;
 import org.apache.brooklyn.core.entity.Entities;
 import org.apache.brooklyn.core.entity.EntityInternal;
 import org.apache.brooklyn.core.entity.lifecycle.Lifecycle;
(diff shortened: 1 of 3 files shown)


tbouron:
@aledsage Comment addressed

software/webapp/src/main/java/org/apache/brooklyn/entity/proxy/AbstractControllerImpl.java
```diff
@@ -317,6 +331,8 @@ protected void computePortsAndUrls() {
 
         sensors().set(PROTOCOL, inferProtocol());
         sensors().set(MAIN_URI, URI.create(inferUrl()));
+        sensors().set(MAIN_URI_MAPPED_SUBNET, URI.create("http://" + sensors().get(SUBNET_ADDRESS) + ":" + sensors().get(PROXY_HTTP_PORT) + "/"));
```
tbouron: That's why I was calling `inferUrl` as it does all this

aledsage: Yes, except it uses the slightly weird `Machines. findSubnetHostname`. On balance, we should probably just live with that, so revert it to calling inferUrl!

To get the unit test to work, we'd need to switch to a location type that implemented `HasSubnetHostname` (annoying - find to leave as a todo, to improve the test in the future).

software/webapp/src/main/java/org/apache/brooklyn/entity/proxy/AbstractControllerImpl.java
```diff
@@ -317,6 +331,8 @@ protected void computePortsAndUrls() {
...
+        sensors().set(MAIN_URI_MAPPED_PUBLIC, URI.create(inferUrlPublic()));
```
neykov: Shouldn't we use `On{Public,Private}NetworkEnricher` instead?

tbouron: @neykov Will the `On{Public,Private}NetworkEnricher` returns the right value? The changes I made was an attempt to follow @ahgittin's proposal

aledsage: @neykov a very good point, but for `AbstractController` it is unfortunately a special case. It wants to use the `domain` sensor for the hostname in the URL, if set. This was added for `GeoscalingDnsService` (and is also useful for things like AWS ELB).

We'd still like to advertise `MAIN_URI_MAPPED_PUBLIC`, but not with the "normal" logic. The same goes I think for `MAIN_URI_MAPPED_SUBNET`: the machine's subnet-address doesn't make sense if it's an AWS ELB that is being used. We want `MAIN_URI_MAPPED_SUBNET` to always be populated, so that other blueprints can use this without worrying about the type of load balancer.

This is probably worth further discussion. But in the mean time I suggest we merge this PR, so we get it working (important for the tutorials/examples that show a simple web-cluster being deployed to GCE and/or Azure).

I'll also look separately at improving `OnPublicNetworkEnricher` so that it can read the public IP from a sensor (rather than only working if the public mapping is registered in `PortForwardManager`, as is done for DNAT).

tbouron: @neykov Interesting approach, I didn't know about those enrichers. I did a quick test and the  `OnSubnetNetworkEnricher` works whereas `OnPublicNetworkEnricher` does as @aledsage explained.

Although, using enrichers would probably be a cleaner solution, we can go back to it later.

aledsage:
Thanks @tbouron - merging now.

PR #68, opened 2016-10-20
BindDns improvements for CentOS

sjcorbett:
On CentOS bind is run by `named`, on Debian it is `bind`. Previously we used the latter in both cases.


neykov:
LGTM, merging.

PR #69, opened 2016-10-28
Update default Zookeeper version and fix NPE

tbouron:
This updates the Zookeeper version to use (version 3.4.5 not available anymore).
It also fixes a NPE on customize phase, caused by:
- the `ZookeeperNode.MY_ID` not set for a single `ZookeeperNode`
- getZookeeperServers() expecting the parent to be and entity of type `ZookeeperEnsemble`, which is not the case for a single `ZookeeperNode`


neykov:
LGTM, merging.

PR #71, opened 2016-11-01
Don't allow multiple context entities in a task's tags

neykov:
Depends on https://github.com/apache/brooklyn-server/pull/385


aledsage:
LGTM; merging.

PR #72, opened 2016-11-01
Fix test to work on any combination of OS & eol type.

neykov:
Change line endings for generated script as it's most likely to run in a unix environment.
Normalize line endings when comparing scripts to work with any combination of OS & line endings (for tests running on Windows mostly). On Windows you can have both CRLF or LF line endings in the files (or even mixed), depending on how git is configured. That's out of our control so try to support any environment.

The recommended confgiuration is "git --global core.autocrlf true", but Apache Jenkins machines have "core.autocrlf" set to "input". The former will lead to CRLF line endings, the latter to LF line endings on checkout.


neykov:
/cc @aledsage as discussed a while ago.
The Windows build machines in Apache Infra have been replaced and the new ones set `core.autocrlf=input` globally which results in most files having LF even on Windows. Can't override it per repo, just for our builds.

googlielmo:
LGTM

sjcorbett:
Thanks @neykov and @googlielmo.

PR #75, opened 2016-11-15
ZooKeeper ensemble fixes

sjcorbett:
* Fixes ZooKeeper ensembles by ensuring that each node has a different id. 
* Improves live tests to test that data can be written and read back.
* ZooKeeperNodes publish their host and port on main.uri
* ZooKeeperEnsemble aggregates its member's main.uris into a list and combines them into a comma-separated string at `zookeeper.endpoints`.

### DIFF
--- a/software/messaging/pom.xml
+++ b/software/messaging/pom.xml
@@ -219,6 +219,10 @@
                     <groupId>storm</groupId>
                     <artifactId>carbonite</artifactId>
                 </exclusion>
+                <exclusion>
+                    <groupId>org.slf4j</groupId>
+                    <artifactId>log4j-over-slf4j</artifactId>
+                </exclusion>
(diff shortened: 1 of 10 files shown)


sjcorbett:
These changes depend on https://github.com/apache/brooklyn-server/pull/439.

sjcorbett:
The build failed because it's missing the dependencies mentioned in brooklyn-server.

drigodwin:
Looks good & tested on AWS. We probably also need to extend the tests at [brooklyn-zookeeper-ensemble](https://github.com/brooklyncentral/brooklyn-zookeeper-ensemble/) at some point to better identify issues.

neykov:
Thanks for reviewing and testing @drigodwin. Merging.

PR #76, opened 2016-11-25
Tweak to password change effector so that sed identifier is escaped

m4rkmckenna:
*Previously*
`s@<regex>@<replace>@g` would not work for a password with an `@`
*Now*
`s/<regex>/<replace>/g` `/` values are escaped in the replace value

### DIFF
--- a/software/database/src/main/java/org/apache/brooklyn/entity/database/mysql/MySqlNodeEffectors.java
+++ b/software/database/src/main/java/org/apache/brooklyn/entity/database/mysql/MySqlNodeEffectors.java
@@ -75,7 +75,7 @@ public class MySqlNodeEffectors {
             DynamicTasks.queue(
                     SshTasks.newSshExecTaskFactory(machine, 
                                     "cd "+entity().getAttribute(MySqlNode.RUN_DIR),
-                                    "sed -i'' -e 's@^\\(\\s*password\\s*=\\s*\\).*$@\\1" + newPass.replace("\\", "\\\\") + "@g' mymysql.cnf")
+                                    "sed -i'' -e 's/^\\(\\s*password\\s*=\\s*\\).*$/\\1" + newPass.replace("\\", "\\\\").replace("/","\\/") + "/g' mymysql.cnf")
                             .requiringExitCodeZero()
                             .summary("Change root password"));
             return null;


geomacy:
👍  LGTM, tested the sed commands generated by the above with values `te\sty` and `te/sty` and both are handled correctly:
```
vagrant@byon5:~$ sed -i'' -e 's/^\(\s*password\s*=\s*\).*$/\1te\\sty/g' mymysql.cnf
vagrant@byon5:~$ 
vagrant@byon5:~$ 
vagrant@byon5:~$ cat mymysql.cnf 
password = te\sty

vagrant@byon5:~$ sed -i'' -e 's/^\(\s*password\s*=\s*\).*$/\1te\/sty/g' mymysql.cnf
vagrant@byon5:~$ 
vagrant@byon5:~$ cat mymysql.cnf 
password = te/sty
```

neykov:
+1 makes sense
Still can't explain to myself why I used the `@` delimited, but then didn't escape the password for it. I had a good reason not to use `/` but can't remember now.

Looking at @geomacy's exmples above I think adding single quote to the escape list is needed as well. Something like:
```
sed -i'' -e 's/^\(\s*password\s*=\s*\).*$/\1te'sty/g' mymysql.cnf
```
where the new password is `te'sty` will break it as well.

A new line is another potential failure point. But I'm fine not coding around this one.

Taking a step back, is `sed` the only option for updating the password? Can we delete the line (with sed) and then append it by other means safely? The only way I could think of dealing with random strings and not going through escape hell is base64'ing it so it goes through bash unaffected. Something like:
```
cat > new_password.tmp <<EOF
base64'ed password from java
EOF
echo -n "password=" >> mysql.cnf
base64 --decode new_password.tmp >> mysql.cnf
```

geomacy:
Good catch @drigodwin I was only testing the sed

m4rkmckenna:
@neykov IMO sed is not the correct tool.

 I know if i was writing a blueprint that was touching any config files id just install [Augeas](http://augeas.net/) and use it. 

I really think brooklyn should be using this for any edit

geomacy:
+1 for Svet's suggestion about encoding it in some way - far better than trying to catch all the possible cases.

tbouron:
@m4rkmckenna `Augeas` seems like a idea. Is it possible to update this blueprint to use this tool?

m4rkmckenna:
@tbouron It would be better to create a pure yaml implementation of this blueprint

PR #78, opened 2016-11-25
Zookeeper node main URI must contain a protocol

m4rkmckenna:
Zookeeper node fails as a `URISyntaxException` is thrown during creation of the main uri


aledsage:
LGTM; merging.
(p.s. https://issues.apache.org/jira/browse/ZOOKEEPER-390 contains a little bit of info about the `zk` url scheme.)

sjcorbett:
https://github.com/apache/brooklyn-library/pull/79 is a follow up to this.

PR #80, opened 2016-12-02
Update MariaDB default version

bostko:
On the latest CentOS 7.2.x MariaDB application fails to start.
This is fixed with the latest MariaDB 5.5.53 so upgrading to it.

Longer term I think it is better if the blueprint is changed to use the yum mariadb repository and to retrieve an rpm from there.


neykov:
LGTM

aledsage:
+1; merging now.

PR #82, opened 2016-12-11
Remove YAML tags from CM output

neykov:
What "adaptForSaltYamlTypes" does is pass a string to the "java.lang.String" constructor. No point in that, use the argument string directly.


geomacy:
👍  looks good and doesn't break org.apache.brooklyn.entity.cm.salt.HighstateTest, which includes use of the "!!python/unicode" form; merging.

PR #83, opened 2017-01-06
Refactor LoadTest

aledsage:
- Add AbstractLoadTest (including small test to sanity-check that test actually works!)
- Change to use yams-based blueprints
- Adds SimulatedVanillaSoftwareProcessImpl
- Adds SimulatedExternalMonitor (for setting sensors on entities)

These changes don't try to fix or investigate any performance/scalability issues etc - they just try to make the tests work better, and be more representative of real-world blueprints (e.g. be written in YAML, and use `VanillaSoftwareProcess`).

Requires that https://github.com/apache/brooklyn-server/pull/506 be merged first.


neykov:
Thanks @aledsage, merging.

PR #84, opened 2017-01-16
Updates zookeeper download URL

m4rkmckenna:
Fastbull is not longer a valid download URL for zookeeper ... changed to apache hosted

### DIFF
--- a/software/messaging/src/main/java/org/apache/brooklyn/entity/zookeeper/ZooKeeperNode.java
+++ b/software/messaging/src/main/java/org/apache/brooklyn/entity/zookeeper/ZooKeeperNode.java
@@ -46,7 +46,7 @@ public interface ZooKeeperNode extends SoftwareProcess {
 
     @SetFromFlag("downloadUrl")
     AttributeSensorAndConfigKey<String, String> DOWNLOAD_URL = ConfigKeys.newSensorAndConfigKeyWithDefault(SoftwareProcess.DOWNLOAD_URL,
-            "http://apache.fastbull.org/zookeeper/zookeeper-${version}/zookeeper-${version}.tar.gz");
+            "http://apache.org/dyn/closer.cgi?filename=zookeeper/zookeeper-${version}/zookeeper-${version}.tar.gz&action=download");
 
     @SetFromFlag("zookeeperPort")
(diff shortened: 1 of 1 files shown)


drigodwin:
Looks good, thanks @m4rkmckenna

software/messaging/src/main/java/org/apache/brooklyn/entity/zookeeper/ZooKeeperNode.java
```diff
@@ -46,7 +46,7 @@
 
     @SetFromFlag("downloadUrl")
     AttributeSensorAndConfigKey<String, String> DOWNLOAD_URL = ConfigKeys.newSensorAndConfigKeyWithDefault(SoftwareProcess.DOWNLOAD_URL,
-            "http://apache.fastbull.org/zookeeper/zookeeper-${version}/zookeeper-${version}.tar.gz");
+            "http://apache.org/dist/zookeeper/zookeeper-${version}/zookeeper-${version}.tar.gz");
```
neykov: From `http://apache.org/dist/`:

`Please do not download from apache.org! If you are currently at apache.org and would like to browse, please visit a nearby mirror site instead.`

Would be fine though if we add the binary to cloudsoft's mirror - it's queried first, right?

m4rkmckenna: @neykov Mirrors come and go where as this site will always be present ... I think it is a smart default

The reason this has broke is because it previously relied on a thirdparty

drigodwin: Do we not want to use: `www.apache.org/dyn/closer.cgi?filename=zookeeper/zookeeper-3.4.9/zookeeper-3.4.9.tar.gz&action=download` as described [here](http://www.apache.org/dev/release-download-pages#closer)

neykov: Agree, but we should avoid putting load on apache.org if possible, by mirroring the binary in cloudsoft.com.

neykov: Great suggestion @drigodwin.

m4rkmckenna: awesome @drigodwin ill amend now

PR #85, opened 2017-01-16
Amends ZK download url so that filename param comes last

m4rkmckenna:
Previously this caused the script to fork / never install


drigodwin:
Thanks for that @m4rkmckenna,  looks good to me

PR #92, opened 2017-02-16
Updates couchDBNode to support V2.0+

drigodwin:
Updates couchDBNode to support CouchDB 2.0+

### DIFF
--- a/software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/couchdb/CouchDBNodeSshDriver.java
+++ b/software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/couchdb/CouchDBNodeSshDriver.java
@@ -79,6 +79,11 @@ public class CouchDBNodeSshDriver extends AbstractSoftwareProcessSshDriver imple
 
     public String getErlangVersion() { return entity.getConfig(CouchDBNode.ERLANG_VERSION); }
 
+    protected boolean isV2() {
+        String version = getVersion();
+        return version.startsWith("2.");
+    }
+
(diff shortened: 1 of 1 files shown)


software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/couchdb/CouchDBNodeSshDriver.java
```diff
@@ -198,24 +211,36 @@ public void customize() {
     @Override
     public void launch() {
         log.info("Launching  {}", entity);
-        newScript(MutableMap.of(USE_PID_FILE, false), LAUNCHING)
-                .body.append(String.format("nohup ./bin/couchdb -p %s -a %s -o couchdb-console.log -e couchdb-error.log -b &", getPidFile(), Os.mergePathsUnix(getRunDir(), getCouchDBConfigFileName())))
-                .execute();
+        String couchDBPath = isV2()? getExpandedInstallDir()+"/rel/couchdb/bin/couchdb" : "./bin/couchdb";
+        ScriptHelper script = newScript(MutableMap.of(USE_PID_FILE, false), LAUNCHING);
```
duncangrant: you could set USE_PID_FILE to isV2()

drigodwin: `USE_PID_FILE` would write it to a `pid.txt` whereas this expects a `couchdb.pid`. I could swap it but I think it might break the 1.x version.

PR #94, opened 2017-03-17
Add entity/config test via REST using redis as an example

ahgittin:
Move the test by @sjcorbett added to brooklyn-server in d7cbaeed46d7fedf8e49df015baa812352060f45
here so it can use library items, as discussed in https://github.com/apache/brooklyn-server/pull/594


sjcorbett:
Looks good.

PR #95, opened 2017-04-07
Fix executeScript effector for MariaDB and PostreSQL

iyovcheva:
Fixes executeScript effector to fail on wrong command.

For postresql the script was returning exit code 0 and required command modification.


geomacy:
LGTM, merging.

PR #99, opened 2017-04-13
Update random passwords

drigodwin:
`Strings.makeRandomId` uses a insecure random so we shouldn't be using it for passwords, I've swapped it to `Identifiers.makeRandomPassword` which does use secure random.


bostko:
LGTM!

PR #104, opened 2017-04-27
Zookeeper now uses subnet for leadership election

Graeme-Miller:
Tested on: GCE, Softlayer, Azure


Graeme-Miller:
Also tested on openstack

Graeme-Miller:
And to complete the set, it has been tested on AWS

PR #106, opened 2017-05-10
Fix GeoDNS tests

drigodwin:
The GEODNS tests are currently failing:
```
 2017-05-10 03:59:21,695 ERROR Problem refreshing group membership: org.apache.brooklyn.util.exceptions.RuntimeInterruptedException: java.lang.InterruptedException
org.apache.brooklyn.util.exceptions.RuntimeInterruptedException: java.lang.InterruptedException
	at org.apache.brooklyn.util.exceptions.Exceptions.propagate(Exceptions.java:124) ~[brooklyn-utils-common-0.12.0-SNAPSHOT.jar:0.12.0-SNAPSHOT]
Caused by: java.lang.InterruptedException: null
	at java.util.concurrent.FutureTask.awaitDone(FutureTask.java:404) [na:1.8.0_121]
Tests run: 42, Failures: 1, Errors: 0, Skipped: 22, Time elapsed: 57.834 sec <<< FAILURE! - in TestSuite
tearDown(org.apache.brooklyn.entity.dns.AbstractGeoDnsServiceTest)  Time elapsed: 31.464 sec  <<< FAILURE!
org.testng.internal.thread.ThreadTimeoutException: Method org.apache.brooklyn.core.test.BrooklynMgmtUnitTestSupport.tearDown() didn't finish within the time-out 30000


Results :

Failed tests: 
  AbstractGeoDnsServiceTest>BrooklynMgmtUnitTestSupport.tearDown:58->BrooklynMgmtUnitTestSupport.destroyManagementContextSafely:71 » ThreadTimeout
```
This swaps them to integration tests.

PR #108, opened 2017-05-11
[WIP] Put back GeoDNS tests

drigodwin:
As @neykov noticed in #106, the failure of these tests could be down to a deadlock fixed in apache/brooklyn-server#671. This puts the tests back.


neykov:
Build is successful. Merging.
Thanks for the changes @drigodwin.

PR #118, opened 2017-06-19
Delete use of deprecated ApplicationBuilder (in tests)

aledsage:
Required updating lots of tests, often to extend `BrooklynAppLiveTestSupport` or `BrooklynAppUnitTestSupport`.

The integration and live tests changed here are used very rarely. It's worth a separate review to decide what we're going to do with them long-term!


ahgittin:
LGTM

PR #124, opened 2017-09-08
use external config provider for demo passwords

ahgittin:
follow-on from https://github.com/apache/brooklyn-server/pull/812

### DIFF
--- a/examples/simple-web-cluster/src/main/resources/visitors-creation-script.sql
+++ b/examples/simple-web-cluster/src/main/resources/visitors-creation-script.sql
@@ -19,17 +19,21 @@
 create database visitors;
 use visitors;
 
-# not necessary to create user if we grant (and not supported in some dialects)
-# create user 'brooklyn' identified by 'br00k11n';
 
-grant usage on *.* to 'brooklyn'@'%' identified by 'br00k11n';
-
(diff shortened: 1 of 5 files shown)


examples/simple-web-cluster/src/main/resources/visitors-creation-script.sql
```diff
@@ -19,18 +19,22 @@
...
-# not necessary to create user if we grant (and not supported in some dialects)
-# create user 'brooklyn' identified by 'br00k11n';
-
-grant usage on *.* to 'brooklyn'@'%' identified by 'br00k11n';
-
-# ''@localhost is sometimes set up, overriding brooklyn@'%', so do a second explicit grant
-grant usage on *.* to 'brooklyn'@'localhost' identified by 'br00k11n';
 
+# the below will create user (and note create user not supported in some dialects)
+grant usage on *.* to 'brooklyn'@'%' identified by '${config["creation.script.password"]}';
```
drigodwin: As I mentioned on the other PR, this is script is used quite a few different places [here](https://brooklyn.apache.org/v/latest/blueprints/multiple-services.html) for example. There should probably be a default so it doesn't break those places.

PR #125, opened 2017-09-13
Improve config and entity descriptions

aledsage:


aledsage:
Thanks @m4rkmckenna - merging now.

PR #138, opened 2017-11-06
add maven-bundle-plugin configuration to allow rebind 

robertgmoss:
(depends on brooklyn-server#873)


robertgmoss:
closing in favour of adding higher up the pom hierarchy

PR #141, opened 2017-11-08
Mysql update password

drigodwin:
If you set the `mysql.datadir` in order to specify an existing database, the MySQL entity still tries to set the root password from blank assuming that it's not set. If the existing database has a password set, even if you define `mysql.password`, it errors. This change tries both blank password and the specified password, fixing this problem.

### DIFF
--- a/software/database/src/main/java/org/apache/brooklyn/entity/database/mysql/MySqlSshDriver.java
+++ b/software/database/src/main/java/org/apache/brooklyn/entity/database/mysql/MySqlSshDriver.java
@@ -186,6 +186,12 @@ public class MySqlSshDriver extends AbstractSoftwareProcessSshDriver implements 
                 boolean hasCreationScript = copyDatabaseCreationScript();
                 timer.waitForExpiryUnchecked();
 
+                /*
+                 * TODO: this should not assume that the password is blank.
+                 * There may be an existing setup specified by datadir which has an existing password. The changePassword function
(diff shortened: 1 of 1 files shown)


drigodwin:
I've tested this by starting this blueprint:
```
brooklyn.catalog:
  version: "1.0.0-SNAPSHOT"
  items:
  - id: test
    version: "1.0.0-SNAPSHOT"
    itemType: template
    license: Apache-2.0
    item:
      services:
      - type: org.apache.brooklyn.entity.software.base.VanillaSoftwareProcess
        brooklyn.config:
          checkRunning.command: |
            echo true
```
Then adding this child:
```
services:
    id: db2
    name: MySQL node 2
    brooklyn.config:
      mysql.password: s0m3rand0mp4ss
      mysql.datadir: /tmp
      datastore.creation.script.url: https://github.com/apache/brooklyn-library/raw/master/examples/simple-web-cluster/src/main/resources/visitors-creation-script.sql
```
Then stopping that child once installed and adding this child:
```
services:
    id: db2
    name: MySQL node 2
    brooklyn.config:
      mysql.password: s0m3rand0mp4ss
      mysql.datadir: /tmp
```
Everything starts and runs as expected

drigodwin:
Thanks @geomacy, I agree about the TODO. The shortcomings of the MySQL entity have already been identified in PRs such as #76 and I think a re-write in YAML is the best course of action. In this case, this addresses a specific blocking bug whilst avoiding changing the functionality and risking it breaking in other places.

PR #144, opened 2017-11-25
BROOKLYN-566: fix SaltEntity rebind to historic state

aledsage:
Fixes https://issues.apache.org/jira/browse/BROOKLYN-566

This class was deleted in https://github.com/apache/brooklyn-library/pull/123, when converting the entity to be a `SoftwareProcess`.

Also see https://github.com/apache/brooklyn-server/pull/905


geomacy:
LGTM, merging

PR #147, opened 2018-02-14
Adds AbstractController changeServerPool

aledsage:
Builds on https://github.com/apache/brooklyn-library/pull/146, which contains the first two comits - please review/merge that first.

Purpose of the new `changeServerPool` effector is to switch the group of target servers, and thus reconfigure the load balancer in one step.

This would be useful for example in blue-green upgrades: the load balancer could start by pointing at the "blue" cluster, and could then be switched to point at the "green" cluster.

### DIFF
--- a/software/webapp/src/main/java/org/apache/brooklyn/entity/proxy/AbstractControllerImpl.java
+++ b/software/webapp/src/main/java/org/apache/brooklyn/entity/proxy/AbstractControllerImpl.java
@@ -33,6 +33,7 @@ import org.apache.brooklyn.api.mgmt.Task;
 import org.apache.brooklyn.api.policy.Policy;
 import org.apache.brooklyn.api.policy.PolicySpec;
 import org.apache.brooklyn.api.sensor.AttributeSensor;
+import org.apache.brooklyn.core.annotation.EffectorParam;
 import org.apache.brooklyn.core.entity.Attributes;
 import org.apache.brooklyn.core.entity.Entities;
 import org.apache.brooklyn.core.entity.EntityInternal;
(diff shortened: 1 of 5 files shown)


software/webapp/src/test/java/org/apache/brooklyn/entity/proxy/AbstractAbstractControllerTest.java
```diff
@@ -0,0 +1,340 @@
...
+import org.testng.annotations.Test;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Lists;
+
+/**
+ * Sub-classes are concrete tests of {@link AbstractController} and {@link AbstractNonProvisionedController},
+ * hence the weird double-abstract name!
```
geomacy: Come to think of it `GenericAbstractController` might be better, or at least look less odd. But no matter...

software/webapp/src/main/java/org/apache/brooklyn/entity/proxy/AbstractControllerImpl.java
```diff
@@ -425,6 +443,20 @@ public void update() {
...
 
+    @Override
+    public void changeServerPool(String groupId) {
+        Group newGroup = (Group) getManagementContext().getEntityManager().getEntity(groupId);
+        if (newGroup == null) {
+            throw new IllegalArgumentException("Group '"+groupId+"' not found");
+        }
+        
+        config().set(SERVER_POOL, newGroup);
+        if (hasServerPoolMemberTrackingPolicy()) {
```
geomacy: The behaviour here is a bit odd; if `serverPoolMemberTrackerPolicy` is `null`, then `hasServerPoolMemberTrackingPolicy` will iterate the `policies()` and if it finds a suitable policy will set it to `serverPoolMemberTrackerPolicy` and return true;  so `addServerPoolMemberTrackingPolicy` will be called, which will promptly remove the `serverPoolMemberTrackerPolicy` and do another iteration through the `policies()`, so finding (presumably) the same one as just before and setting it again.

aledsage: What happens is...
If we don't already have a serverPoolMemberTrackingPolicy (e.g. because we haven't been started yet), then keep it that way (i.e. don't add one).

In `hasServerPoolMemberTrackingPolicy`, it handles the case where the field has not been set - it is effectively transient because fields are not persisted (only attributes, config and adjuncts).

When we then call `addServerPoolMemberTrackingPolicy`, it's into the code I didn't touch. Agreed that method is a bit weird - it'll remove it if `serverPoolMemberTrackerPolicy` is not null, but won't if a policy of that type already exists with the field being null (which presumably could happen on rebind).

In conclusion, I agree that `addServerPoolMemberTrackingPolicy` is a bit weird. But I'm ok with the logic behind the code I've changed/added :-)

software/webapp/src/test/java/org/apache/brooklyn/entity/proxy/AbstractAbstractControllerTest.java
aledsage: Agreed - but because it's only for a test I'm not too bothered about the naming.

aledsage:
Rebased against master (so only the final commit was present in the PR), and then merged.

PR #148, opened 2018-03-08
Fix issue with JBoss 7 installation

tbouron:
The `jboss-modules.jar` that comes with jBoss 7.1.1 is of version `1.1.1.GA`. However, this has issue on initialisation when one launches the server.

Hopefully, version `1.1.5.GA` fixes the init issue. As jBoss 7.1.1 is EOF, we need to download and replace
this jar before starting up the server.


geomacy:
JBoss from `master` fails to start but starts successfully with this PR. Will merge.

PR #149, opened 2018-03-09
Deprecate JBoss 7 entity

tbouron:
As the title, this deprecates JBoss 7, mainly because it is EOF.

See discussion here: https://lists.apache.org/thread.html/26d0225c8608fb398c2ea7b2d097a8a8784710c3bc7ea86270da28b4@%3Cdev.brooklyn.apache.org%3E


tbouron:
@geomacy Done!

PR #153, opened 2018-05-01
Add Jenkinsfile for CI build

tbouron:
This capture the current Jenkins configuration, so that is repeatable and documented.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,82 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,87 @@
...
+            // Send email notifications
+            step([
+                $class: 'Mailer',
+                notifyEveryUnstableBuild: true,
+                recipients: 'dev@brooklyn.apache.org',
+                sendToIndividuals: false
+            ])
+        }
+
+        // TODO: Build full brooklyn project
```
geomacy: Does this need to be in here? Surely the trigger will be built into the top level brooklyn build job?  I'd have thought it's not up to upstream jobs like this to decide what downstream jobs need to be triggered, rather it's up to any number of downstreams to see this job succeeding and then decide to rebuild themselves?

Jenkinsfile
```diff
@@ -0,0 +1,87 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+node(label: 'ubuntu') {
+    catchError {
+        def environmentDockerImage
+
+        def dockerTag = env.BUILD_TAG.replace('%2F', '-')
```
geomacy: why's this needed?  The tags have the form `jenkins-brooklyn-library-master-docker-179`

tbouron: @geomacy this is needed because if the job name contains a `/` (which is possible as the name will be the actual branch to build) it will be replaced by the URL escaped version, i.e. `%2F`. However, docker tag accepts only alphanumerical characters + `.`, `-` and `_`

geomacy: btw `%2F` is `/`, but that's what you intend here, I guess? Just checking.

Jenkinsfile
tbouron: Ah you are right @geomacy I can use `pipelineTriggers` into the downstream project. I'll remove the TODO then

PR #159, opened 2018-09-24
Jenkins docker: use non-root user

aledsage:
Implementing part of the advice from infra in https://issues.apache.org/jira/browse/INFRA-16417

---
They advised we use `-u 910:910`, rather than the mvn command running as root in the container. 

Note that running the docker build on my local (mac) laptop, the user ids are interesting! In the container, (which use bind mounts for `.m2` and and the workspace) are owned by root:root. However, on my laptop, the files created are still owned by my own user.

Magic?! I'm therefore not sure whether this change will make much difference.

---
The other big change we need (not done here - I'll look at that next), recommended in INFRA-16417, is to STOP bind mounting .m2. They said: "Please don't use a bind mount for filesystems you intend to write to".

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -36,16 +36,16 @@ node(label: 'ubuntu') {
             }
 
             stage('Run tests') {
-                environmentDockerImage.inside('-i --name brooklyn-${DOCKER_TAG} -v ${WORKSPACE}/.m2:/root/.m2 -v ${WORKSPACE}:/usr/build -w /usr/build') {
-                    sh 'mvn clean install'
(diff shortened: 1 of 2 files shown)


tbouron:
LGTM and follow what docs for the maven image: https://hub.docker.com/_/maven/ "Running as non-root"

As Jenkins seems happy, I'm happy and I'll merge this. Thanks @aledsage

tbouron:
One thing though, could you port the change to the other submodule @aledsage ?

aledsage:
@tbouron will do, but will first confirm that brooklyn-library master build works, and will also look to see if/how we can avoid the bind mount of the writable directories.

PR #163, opened 2018-10-16
Use official apache archive URL to download tomcat libraries

tbouron:
The previous domain has completely changed and not available anymore


asfgit:
Can one of the admins verify this patch?

PR #172, opened 2019-04-24
Postgresql version change from 9.3 to 10 

frogfather:
https://www.pivotaltracker.com/story/show/165526205
brooklyn-postgresql-java test was failing due to incorrect download url. 
Since ver 9.3 is no longer maintained it was not possible to obtain a current url for Centos 7 so the version was changed to 10.
PostgreSql tests passing, however Integration tests did not work even before this change.


asfgit:
Can one of the admins verify this patch?

PR #180, opened 2020-06-01
corrects broken composer palette icon for Artemis

frogfather:
The Artemis logo on the composer palette was a hotlink from the ActiveMQ site. The link was no longer valid. Added static image to brooklyn-library and adjusted path.


duncangrant:
I guess we should use the same log for all the activemq entites?  So might be better to update the existing one?

frogfather:
> I guess we should use the same log for all the activemq entites? So might be better to update the existing one?

Done!

ahgittin:
I'd be inclined to remove `artemis-logo.png` and simply reference the updated activemq for that one.

@frogfather do you have an Apache ICLA?

i'll update the notice metadata to note we are including that file in our source dist.

ahgittin:
looks good.  i've added a comment to our notices about use of logos, in a separate PR.

as soon as @frogfather has an ICLA approved i think this can be merged

ahgittin:
@frogfather you're listed with an ICLA, so merging

PR #184, opened 2021-09-03
Added Tomcat9 type and modified logos to see the difference between b…

iuliana:
Added Tomcat9 type and modified logos to see the difference between Brooklyn Tomcat types. Previouslt it was difficult to discern between version 7 and 8. And now that we added Tomcat 9, thigs would have been even more confuzing.
Also upgraded old version of Tomcats.


jcabrerizo (commented):
I like it!

PR #191, opened 2022-07-20
Fixed API for KnifeTaskFactory by replacing guava.Function with java.…

iuliana:
Fixed API for KnifeTaskFactory by replacing `com.google.common.base.Function` with `java.util.function.Function`.


ahgittin:
LGTM

